PR #13, opened 2019-10-11
refactor(protocol): Add fixed codec

mkxbl:
**What type of PR is this?**
refactor

add fixed codec to protocol

Fixes #


yejiayu:
Don't implement `FixedCodec` for `Tx Approve`, because we don't need it in the executor2.0 design.

protocol/src/fixed_codec/epoch.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        Ok(Proof {
+            epoch_id,
+            round,
+            epoch_hash,
+            signature,
+            bitmap
+        })
+    }
+}
```
KaoImin: The implement of `rlp::Decodable` may refer to this https://github.com/cryptape/overlord/blob/master/src/codec.rs#L19:L34

protocol/src/fixed_codec/tests/mod.rs
```diff
@@ -0,0 +1,269 @@
+extern crate test;
+
+mod fixed_codec;
+
+use bytes::Bytes;
+use num_traits::FromPrimitive;
+use rand::random;
+use std::collections::BTreeMap;
```
KaoImin: Like this:
```rust
use std::collections::BTreeMap;

use bytes::Bytes;
use num_traits::FromPrimitive;
use rand::random;
```

mkxbl:
> Don't implement `FixedCodec` for `Tx Approve`, because we don't need it in the executor2.0 design.

ok

mkxbl:
> Needs `cargo fmt`

not finished yet

yejiayu:
@zeroqn PTAL

protocol/src/fixed_codec/receipt.rs
```diff
@@ -0,0 +1,150 @@
...
+        })
+    }
+}
+
+const TRANSFER_RESULT_FLAG: u8 = 0;
+const DEPLOY_RESULT_FLAG: u8 = 1;
+const CALL_RESULT_FLAG: u8 = 2;
+const FAIL_RESULT_FLAG: u8 = 3;
+
+#[allow(unused_variables)]
```
yejiayu: Useless drive?

protocol/src/fixed_codec/receipt.rs
```diff
@@ -0,0 +1,150 @@
...
+                    ContractType::Native => 3,
+                };
+                s.append(&type_flag);
+            }
+            ReceiptResult::Call {
+                contract,
+                return_value,
+                logs_bloom,
+            } => {
+                // TODO(@yejiayu): The interface for `call` is about to be modified.
```
yejiayu: Remove this TODO and implement it

protocol/src/fixed_codec/transaction.rs
```diff
@@ -0,0 +1,170 @@
...
+    primitive::{Balance, ContractType, Fee, Hash, UserAddress},
+    transaction::{CarryingAsset, RawTransaction, SignedTransaction, TransactionAction},
+};
+use crate::{impl_default_fixed_codec_for, ProtocolResult};
+
+// Impl ProtocolFixedCodec trait for types
+impl_default_fixed_codec_for!(transaction, [RawTransaction, SignedTransaction]);
+
+impl rlp::Encodable for RawTransaction {
+    fn rlp_append(&self, s: &mut rlp::RlpStream) {
```
yejiayu: Implementing so many nested `RawTransaction` encode can be confusing for the Toolchain.

Maybe we should flatten it? like this
````rs
s.append(TX_FLAG);
s.append(transfer_receiver);
s.append(transfer_amount);
s.append(fee_asset_id);
.....
````

mkxbl: ok. working on it

PR #19, opened 2019-10-16
fix(mempool): Fix concurrent insert bug of mempool

rev-chaos:
**What type of PR is this?**

Fixes #


core/mempool/src/map.rs
```diff
@@ -22,6 +25,7 @@ where
             buckets.push(Bucket {
                 // Allocate enough space to avoid triggering resize.
                 store: RwLock::new(HashMap::with_capacity(cache_size)),
+                lock:  Arc::new(AtomicBool::new(false)),
```
zeroqn: Arc? Is it really needed?

PR #20, opened 2019-10-16
feat(mempool): implement cached batch txs broadcast

zeroqn:
**What type of PR is this?**

feat

Cache batch of signed transactions until we reach specified size or interval, then broadcast them
in one network message.

Fixes #

### DIFF
--- a/core/mempool/src/adapter/mod.rs
+++ b/core/mempool/src/adapter/mod.rs
@@ -142,7 +142,7 @@ impl<G: Gossip + Unpin + Clone + 'static> Future for IntervalTxsBroadcaster<G> {
             let delay = &mut self.as_mut().delay;
             pin_mut!(delay);
 
-            let _ = ready!(delay.poll(ctx));
+            ready!(delay.poll(ctx));
 
             if !self.txs_cache.is_empty() {
                 self.do_broadcast();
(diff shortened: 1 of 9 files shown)


core/mempool/src/adapter/mod.rs
```diff
@@ -25,11 +26,19 @@ use crate::adapter::message::{
 use crate::adapter::rlp_types::RlpRawTransaction;
 use crate::MemPoolError;
 
+pub const DEFAULT_BROADCAST_TXS_SIZE: usize = 20000;
```
yejiayu:
```suggestion
pub const DEFAULT_BROADCAST_TXS_SIZE: usize = 200;
```

core/mempool/src/adapter/mod.rs
```diff
@@ -25,11 +26,19 @@ use crate::adapter::message::{
...
+pub const DEFAULT_BROADCAST_TXS_INTERVAL: u64 = 2; // seconds
```
yejiayu:
```suggestion
pub const DEFAULT_BROADCAST_TXS_INTERVAL: u64 = 200; // ms
```

zeroqn:
> Please add these new configuration items to `devtools/chain/config.toml`

Done

PR #29, opened 2019-10-24
feat: change rlp in executor to fixed-codec

huwenchao:
**What type of PR is this?**

Fixes #


core/executor/src/fixed_types/mod.rs
```diff
@@ -54,67 +48,13 @@ impl FixedAsset {
 
 impl ContractSer for FixedAsset {
```
yejiayu: The `ContractSer` is a useless trait,  any struct that implements the `Fixed-codec` can be stored in the `state trie`.

huwenchao: > The `ContractSer` is a useless trait, any struct that implements the `Fixed-codec` can be stored in the `state trie`.

What if there are contract-specific structs  not implement in protocol/types?
e.g., structs in crosschain contract

PR #30, opened 2019-10-25
fix(mempool): Always get the latest epoch id when `package`.

yejiayu:
**What type of PR is this?**
fix

Fixes #

### DIFF
--- a/core/mempool/src/adapter/mod.rs
+++ b/core/mempool/src/adapter/mod.rs
@@ -158,6 +158,7 @@ where
     }
 
     async fn get_latest_epoch_id(&self, _ctx: Context) -> ProtocolResult<u64> {
-        Ok(self.storage.get_latest_epoch().await?.header.epoch_id)
+        let epoch_id = self.storage.get_latest_epoch().await?.header.epoch_id;
+        Ok(epoch_id)
     }
 }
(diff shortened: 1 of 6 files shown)


core/mempool/src/adapter/mod.rs
```diff
@@ -156,4 +156,8 @@ where
             }
         }
     }
+
+    async fn get_latest_epoch_id(&self, _ctx: Context) -> ProtocolResult<u64> {
+        Ok(self.storage.get_latest_epoch().await?.header.epoch_id)
```
KaoImin:
```suggestion
        let epoch_id = self.storage.get_latest_epoch().await?.header.epoch_id;
        Ok(epoch_id)
```

protocol/src/traits/mempool.rs
```diff
@@ -58,4 +58,6 @@ pub trait MemPoolAdapter: Send + Sync {
     async fn check_transaction(&self, ctx: Context, tx: SignedTransaction) -> ProtocolResult<()>;
 
     async fn check_storage_exist(&self, ctx: Context, tx_hash: Hash) -> ProtocolResult<()>;
+
+    async fn get_latest_epoch_id(&self, _ctx: Context) -> ProtocolResult<u64>;
```
KaoImin: Why the argument is `_ctx`？

PR #33, opened 2019-10-26
doc: auto gen api

homura:
**What type of PR is this?**
doc
Supports auto generating docs

Fixes #

### DIFF
--- a/core/api/src/schema/epoch.rs
+++ b/core/api/src/schema/epoch.rs
@@ -1,26 +1,46 @@
 use crate::schema::{Address, Hash, MerkleRoot, Uint64};
 
 #[derive(GraphQLObject, Clone)]
-#[graphql(description = "Epoch")]
+#[graphql(
+    description = "Epoch is a single digital record created within a blockchain. \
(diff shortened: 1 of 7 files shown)


core/api/src/schema/epoch.rs
```diff
@@ -1,26 +1,46 @@
...
-    pub pre_hash:     Hash,
-    pub timestamp:    Uint64,
-    pub order_root:   MerkleRoot,
+    #[graphql(
+        description = "Identifier of a chain in order to prevent replay attacks across channels "
+    )]
+    pub chain_id: Hash,
+    #[graphql(description = "Known as the block height like other blockchain")]
+    pub epoch_id: Uint64,
+    #[graphql(description = "The merkle root of the previous epoch")]
```
huwenchao:
```suggestion
    #[graphql(description = "Previous epoch hash")]
```

This is a hash, not a mercke root.

homura: OK, I have make the description more clearly now

PR #43, opened 2019-12-01
feat: metrics logger

huwenchao:
**What type of PR is this?**

A logger and metrics implementation.

Implement a crate to make log and add metrics easier.

Check `README.md` for more information.


huwenchao:
usage example: https://github.com/huwenchao/muta/commit/7c04ff3e906ab5e155c8258b63195740f9d1f0f5

yejiayu:
So, what is the format of the output?

common/logger/log.yml
```diff
@@ -0,0 +1,36 @@
+# This file is yaml style config, can make testing the logger more easily.
```
yejiayu: Where is it used?

common/logger/src/lib.rs
```diff
@@ -1,73 +1,90 @@
...
-        trace!("Hello {}", "World!");
-        trace!("Hell{} {}", "o", "World!");
-        trace!("Hell{} World", 0);
-    }
+    let metrics_logger = Logger::builder().additive(false).appender("metrics").build(
+        "metrics",
+        if metrics {
+            LevelFilter::Trace
+        } else {
+            LevelFilter::Off
```
yejiayu: Does this mean that `trace` information is lost?

The PR may be missing some necessary instructions.

huwenchao: > Does this mean that trace logs will be discarded?

No, only the `metrics` target is discarded.

> The PR may be missing some necessary instructions.

That makes sense. I'll add a readme file.

common/logger/log.yml
huwenchao: It is the yaml style of the rust code in `init` function in `lib.rs`, not used here, but makes it easier to understand the code and test new logger config.

If someone wants to have more customized logger config other than that in the `config.toml` file, he/she may need this.

common/logger/Cargo.toml
```diff
@@ -7,5 +7,6 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-env_logger = "0.6"
 log = "0.4"
+log4rs = "0.8"
+serde_json = "1.0"
```
yejiayu: newline

common/logger/log.yml
```diff
@@ -0,0 +1,36 @@
...
+  appenders:
+  - console
+  - file
+
+loggers:
+  metrics:
+    level: trace
+    appenders:
+    - metrics
+    additive: false
```
yejiayu: newline

devtools/chain/config.toml
```diff
@@ -41,3 +41,11 @@ precommit_denominator = 30
...
 [executor]
 light = false
+
+[logger]
+filter = "info"
+log_to_console = true
+console_show_file_and_line = false
+log_path = "logs/"
+log_to_file = true
+metrics = true
```
yejiayu: newline

common/logger/Cargo.toml
huwenchao: > newline

I wondered why newline is necessary, did some search and found this <https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline>.
 Pretty interesting.

PR #45, opened 2019-12-04
reafctor(consensus): redesign the consensus architecture

KaoImin:
**What type of PR is this?**

refactor

rewrite muta consensus

Fixes #


KaoImin:
This PR has been tested among 4 nodes. The parallel process of consensus and execute, and the synchronous process are both working regularly. The log message will be rewritten after reviewing it.

core/consensus/src/engine.rs
```diff
@@ -199,6 +201,8 @@ impl<Adapter: ConsensusAdapter + 'static> Engine<FixedPill, FixedSignedTxs>
             )
             .await?;
 
+        let mut set = self.exemption_hash.write();
```
yejiayu: There are potential deadlocks and you need to release the locks as soon as possible.

A good practice is to encapsulate clear as a function.

PR #53, opened 2019-12-11
feat: support consensus tracing

homura:
**What type of PR is this?**
feature

The consensus algorithm always plays with a distributed system, debugging or optimizing is so hard. If we can record events to describe what happen when consensus state changes, and then replay with a visualization dashboard, the debugging or optimizing would be easier.

Fixes #

### DIFF
--- a/core/consensus/src/trace.rs
+++ b/core/consensus/src/trace.rs
@@ -33,7 +33,7 @@ impl Trace for MetricTracer {
 }
 
 pub fn init_tracer(address: String) {
-    if let Err(_) = set_boxed_tracer(Box::new(MetricTracer::new(address))) {
+    if set_boxed_tracer(Box::new(MetricTracer::new(address))).is_err() {
         error!("tracing: tracing init failed");
     }
 }
--- a/core/consensus/src/trace.rs
+++ b/core/consensus/src/trace.rs
@@ -1,6 +1,6 @@
 use log::error;
 use moodyblues_sdk::time::now;
-use moodyblues_sdk::trace::{set_boxed_tracer, Metadata, Trace, TracePoint};
+use moodyblues_sdk::trace::{set_boxed_tracer, Metadata, Trace, TracePoint, SetTraceError};
 use serde_json::to_string;
 
 struct MetricTracer {
@@ -32,8 +32,6 @@ impl Trace for MetricTracer {
     }
 }
 
-pub fn init_tracer(address: String) {
-    if set_boxed_tracer(Box::new(MetricTracer::new(address))).is_err() {
-        error!("tracing: tracing init failed");
-    }
+pub fn init_tracer(address: String) -> Result<(), SetTraceError> {
+    set_boxed_tracer(Box::new(MetricTracer::new(address)))
 }
--- a/core/consensus/src/trace.rs
+++ b/core/consensus/src/trace.rs
@@ -1,9 +1,9 @@
-use common_logger::{metrics, object};
+use log::error;
 use moodyblues_sdk::time::now;
-use moodyblues_sdk::trace::{start_epoch, set_boxed_tracer, Metadata, Trace, TracePoint};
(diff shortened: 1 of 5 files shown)


core/consensus/src/trace.rs
```diff
@@ -0,0 +1,39 @@
...
+            address: self.address.clone(),
+        }
+    }
+
+    fn now(&self) -> u64 {
+        now()
+    }
+}
+
+pub fn init_tracer(address: String) {
```
yejiayu: This should return an error.

core/consensus/src/trace.rs
```diff
@@ -33,7 +33,7 @@ impl Trace for MetricTracer {
 }
 
 pub fn init_tracer(address: String) {
-    if let Err(_) = set_boxed_tracer(Box::new(MetricTracer::new(address))) {
+    if set_boxed_tracer(Box::new(MetricTracer::new(address))).is_err() {
```
KaoImin: Eat the error here?

core/consensus/src/trace.rs
```diff
@@ -32,8 +32,6 @@ impl Trace for MetricTracer {
     }
 }
 
-pub fn init_tracer(address: String) {
-    if set_boxed_tracer(Box::new(MetricTracer::new(address))).is_err() {
-        error!("tracing: tracing init failed");
-    }
+pub fn init_tracer(address: String) -> Result<(), SetTraceError> {
```
yejiayu: Return ProtocolResult

src/main.rs
```diff
@@ -266,6 +267,7 @@ async fn start(cfg: &Config) -> ProtocolResult<()> {
         consensus_interval: cfg.consensus.interval,
     };
 
+    init_tracer(my_address.as_hex()).unwrap();
```
KaoImin: Use `init().expect(error message)` here.

core/consensus/src/trace.rs
```diff
@@ -0,0 +1,37 @@
+use log::error;
```
KaoImin:
```suggestion
use log::{error, trace};
```

core/consensus/src/trace.rs
```diff
@@ -0,0 +1,37 @@
...
+impl MetricTracer {
+    fn new(address: String) -> MetricTracer {
+        MetricTracer { address }
+    }
+}
+
+impl Trace for MetricTracer {
+    fn report(&self, point: TracePoint) {
+        match to_string(&point) {
+            Ok(json) => log::trace!(target: "metrics", "{}", json),
```
KaoImin:
```suggestion
            Ok(json) => trace!(target: "metrics", "{}", json),
```

src/main.rs
homura: > Use `init().expect(error message)` here.

returned ProtocolResult now, the `unwrap` is similar to `expect`

core/consensus/src/trace.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    fn now(&self) -> u64 {
+        now()
+    }
+}
+
+pub fn init_tracer(address: String) -> ProtocolResult<()> {
+    set_boxed_tracer(Box::new(MetricTracer::new(address))).map_err(|_| {
+        ProtocolError::from(ConsensusError::Other("failed to init tracer ".to_string()))
```
yejiayu: ConsensusError::Other("failed to init tracer ".to_string().into

homura: map_err is convenient here, why still use `Err into`

yejiayu: `ProtocolError::from` is unnecessary.

Try replacing it with `ConsensusError::Other("failed to init tracer ".to_string()).into`

homura: OK, I know, this looks better

PR #55, opened 2019-12-13
[ᚬframework] feat(core/binding): add StoreMap, StoreArray 

mkxbl:
**What type of PR is this?**
feat(core/binding)
add StoreMap, StoreArray 

Fixes #


core/binding/src/store/map.rs
```diff
@@ -1 +1,163 @@
...
+use bytes::Bytes;
+
+use protocol::fixed_codec::{FixedCodec, FixedCodecError};
+use protocol::traits::{ServiceState, StoreMap};
+use protocol::types::Hash;
+use protocol::ProtocolResult;
+
+use crate::store::StoreError;
+
+pub struct FixedKeys<K: FixedCodec> {
```
yejiayu: It should be moved to `store/mod.rs`

core/binding/src/store/map.rs
```diff
@@ -1 +1,163 @@
...
+    fn insert(&mut self, key: K, value: V) -> ProtocolResult<()> {
+        let mk = self.get_map_key(&key)?;
+        self.state.borrow_mut().insert(mk, value)?;
+
+        if !self.contains(&key)? {
+            self.keys.inner.push(key);
+            self.state
+                .borrow_mut()
+                .insert(self.var_name.clone(), self.keys.encode_fixed()?)
+        } else {
```
yejiayu: Useless else.

I bet you didn't run `cargo clippy --all targets --all features`.

core/binding/src/store/map.rs
```diff
@@ -1 +1,163 @@
...
+            Ok(())
+        }
+    }
+
+    // TODO(@zhounan): Atomicity of insert(k, v) and insert self.keys to
+    // ServiceState is not guaranteed for now That must be settled soon after.
+    fn remove(&mut self, key: &K) -> ProtocolResult<()> {
+        if let true = self.contains(key)? {
```
yejiayu: WTF???

This is bullshit.

yejiayu: I suggest you judge it is not contains, if it does not contain, return an error.

This guarantees as little code indentation as possible.

core/binding/src/store/map.rs
```diff
@@ -1 +1,163 @@
...
+        } else {
+            Err(StoreError::GetNone.into())
+        }
+    }
+
+    fn len(&self) -> ProtocolResult<usize> {
+        Ok(self.keys.inner.len())
+    }
+
+    fn for_each<F>(&mut self, mut f: F) -> ProtocolResult<()>
```
yejiayu: In both `StoreMap` and `StoreArray`, the `for_each` implementations have serious performance issues.

If you don't plan to solve it now, I suggest you mark TODO

core/binding/src/store/map.rs
mkxbl: Don't understand. Can you talk about it more specifically ?

core/binding/src/store/map.rs
mkxbl: Why? Where are the performance issues specifically?

mkxbl:
make clippy is fine
   
ping@yejiayu

core/binding/src/store/map.rs
yejiayu:
```rust
if true == true { } // × shit code
if true { } // √
```

core/binding/src/store/map.rs
yejiayu: It will be re-serialized even if the value has not been modified.

core/binding/src/store/map.rs
mkxbl: fixed

core/binding/src/store/map.rs
mkxbl: TODO marked

mkxbl:
fixed and add TODO comments

PR #59, opened 2019-12-20
feat: add name field to consensus tracing metric log

homura:
**What type of PR is this?**
feat

add name field to consensus tracing metric log to distinguish with other metrics 

Fixes #


core/consensus/src/trace.rs
```diff
@@ -18,9 +18,21 @@ impl MetricTracer {
     }
 }
 
+fn err() -> ProtocolError {
+    ConsensusError::Other("tracing: failed when parse point".to_string()).into()
```
yejiayu: why not implement `Serde::Error` in ConsensusError?

homura: The TracePoint only serialize but not deserialize now, this means that there will be no IO during the processs. So I think error does not occur often

core/consensus/src/trace.rs
```diff
@@ -18,9 +18,21 @@ impl MetricTracer {
...
+}
+
+fn to_trace_str(point: TracePoint) -> ProtocolResult<String> {
+    let mut json = to_value(point).map_err(|_| err())?;
```
huwenchao: It is a bad idea to swallow the original error information here.
There are 3 places you use `err()` function here. If error does happens, the `Err(e) => error!("tracing: convert json error {:?}", e),` only print `tracing: failed when parse point` for you, no other useful information to help debugging.
If you think error should not happen, use except, otherwise try to wrap the original error with necessary information.

homura: OK, I see, thx for the suggestion. I have make the error more detailed

homura:
Seom strange conflict in this PR, so close it and new one in #69

https://github.com/homura/muta/pull/59/commits/f0302586d872be30bc454ed577dfd143b7bded1d
https://github.com/homura/muta/commit/f0302586d872be30bc454ed577dfd143b7bded1d

PR #60, opened 2019-12-21
feat(consensus): develop aggregate crypto with overlord

KaoImin:
**What type of PR is this?**
feature

Develop aggregate crypto with `ophelia-bls-amcl` crypto protocol.
Add some fields to the config file.

Fixes #


devtools/chain/config.toml
```diff
@@ -30,6 +30,9 @@ cycles_limit = 99999999
 cycles_price = 1
 interval = 3000
 verifier_list = [ "10f8389d774afdad8755ef8e629e5a154fddc6325a" ]
+private_key = "000000000000000000000000000000001abd6ffdb44427d9e1fcb6f84e7fe7d98f2b5b205b30a94992ec24d94bb0c970"
```
yejiayu: What's the difference between it and https://github.com/nervosnetwork/muta/blob/9376d0ef9d92fd55a6483f1f76b1b0d603419f9d/devtools/chain/config.toml#L5?

KaoImin: There is no difference, perhaps we should delete the unit test code.

src/main.rs
```diff
@@ -270,6 +273,27 @@ async fn start(cfg: Config) -> ProtocolResult<()> {
         consensus_interval: cfg.consensus.interval,
     }));
 
+    assert!(cfg.consensus.verifier_list.len() == cfg.consensus.public_keys.len());
+    let mut bls_pub_keys = HashMap::new();
+    for item in cfg
```
yejiayu: for (pubkey, bls_pubkey) in cfg

PR #61, opened 2019-12-23
fix(consensus): empty block receipts root

KaoImin:
**What type of PR is this?**
bugfix

* fix a bug cause panic when generating receipts root of an empty block
* fix a bug of update epoch info after commit

Fixes #


devtools/docker/dockercompose/config-bft-1.toml
```diff
@@ -8,12 +8,12 @@ privkey = "136dbd5402e4002d9efb3f67eb8250b719e54813a4369cf8ad26ef916d938850"
 data_path = "./devtools/chain/data"
 
 [graphql]
-listening_address = "0.0.0.0:8000"
+listening_address = "192.168.0.106:8000"
```
yejiayu: miaomiaomiao?

KaoImin: sorry

core/consensus/src/status.rs
```diff
@@ -184,7 +188,7 @@ impl CurrentConsensusStatus {
         }
 
         let len = receipt_roots.len();
-        if self.receipt_root.len() < len || check_vec_roots(&self.receipt_root, receipt_roots) {
+        if self.receipt_root.len() < len || !check_vec_roots(&self.receipt_root, receipt_roots) {
```
huwenchao: The whole function 

```rust
    fn update_receipt_root(&mut self, receipt_roots: &[MerkleRoot]) -> ProtocolResult<()> {
        if receipt_roots.is_empty() {
            return Ok(());
        }

        let len = receipt_roots.len();
        if self.receipt_root.len() < len || !check_vec_roots(&self.receipt_root, receipt_roots) {
            warn!("receipt root: {:?}", receipt_roots);
            return Err(ConsensusError::StatusErr(StatusCacheField::ReceiptRoot).into());
        }

        let tmp = self.receipt_root.split_off(len);
        self.receipt_root = tmp;
        Ok(())
    }
```

can be simplify as below:

```rust
    fn update_receipt_root(&mut self, receipt_roots: &[MerkleRoot]) -> ProtocolResult<()> {
        if !check_vec_roots(&self.receipt_root, receipt_roots) {
            warn!("receipt root: {:?}", receipt_roots);
            return Err(ConsensusError::StatusErr(StatusCacheField::ReceiptRoot).into());
        }

        self.receipt_root = self.receipt_root.split_off(receipt_roots.len());
        Ok(())
    }
```

PR #63, opened 2019-12-23
chore(ci): use sccache and limit its cache size to 2G

zeroqn:
Try fix cahce timeout

**What type of PR is this?**
chore

Fix travis cache timeout again :)
Use sccache and limit its cache size to 2G.

Fixes #


zeroqn:
Under 10 minutes now

PR #71, opened 2019-12-25
[ᚬframework] feat(framework): Service executor.

yejiayu:
**What type of PR is this?**
feature

The PR contains the following two parts:
- Service executor: used to execute the service specified in the transaction. see `framework/src /executor/mod.rs`
- AssetService: a build-in service through which you can create assets and transfer assets. see `built-in-services/asset/src/lib.rs`

Fixes #

The **dispatcher** module used for inter-service invocation is not yet implemented. @mkxbl


build-in-services/asset/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        ctx: Context,
+        payload: CreateAssetPayload,
+    ) -> ProtocolResult<Asset> {
+        let caller = ctx.get_caller();
+        let payload_str =
+            serde_json::to_string(&payload).map_err(|e| ServiceError::JsonParse(e))?;
+
+        let id = Hash::digest(Bytes::from(payload_str + &caller.as_hex()));
+
+        if self.assets.contains(&id)? {
```
zeroqn: No symbol, name check?

build-in-services/asset/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+            return Err(ServiceError::LackOfBalance {
+                expect: value,
+                real:   caller_balance,
+            }
+            .into());
+        }
+
+        let to_balance = self.sdk.get_account_value(&to, &asset_id)?.unwrap_or(0);
+        self.sdk
+            .set_account_value(&to, asset_id.clone(), to_balance + value)?;
```
zeroqn: Should use ```checked_add``` or ```overflowing_add``` or ```saturating_add```.

build-in-services/asset/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        self.sdk
+            .set_account_value(&caller, asset_id.clone(), caller_balance - value)?;
```
zeroqn: Same as above. ```checked_sub``` or ....

zeroqn: IMO, we should sub first, if there's an error, we should handle it, restore user's balance.

yejiayu: Any errors that occur during execution are rolled back.

build-in-services/asset/src/lib.rs
yejiayu: Id is generated by caller + name + symbol

yejiayu:
ping @zeroqn

protocol/Cargo.toml
```diff
@@ -12,19 +12,19 @@ derive_more = "0.15"
...
 lazy_static = "1.4"
 num-bigint = "0.2"
-ethbloom = "0.6"
+ethbloom = "0.8"
 bytes = "0.4"
 hex = "0.3"
 prost = "0.5"
 hasher = { version = "0.1", features = ['hash-keccak'] }
 creep = "0.1"
 bincode = "1.1"
```
KaoImin:
```suggestion
bincode = "1.2"
```

yejiayu:
ping @KaoImin

PR #82, opened 2020-01-01
[ᚬmaster] feat(framework): add service-call-service feature

mkxbl:
**What type of PR is this?**

add framework feature: service call service

using ServiceSDK call another service is supported now
```rust
pub trait ServiceSDK {
    fn read(
        &self,
        ctx: &ServiceContext,
        service: &str,
        method: &str,
        payload: &str,
    ) -> ProtocolResult<String>;

    fn write(
        &mut self,
        ctx: &ServiceContext,
        service: &str,
        method: &str,
        payload: &str,
    ) -> ProtocolResult<String>;

      ... ...
}

```

Fixes #

- path of tests:  framework/src/executor/tests/service_call_service.rs

### DIFF
--- a/framework/src/binding/sdk/mod.rs
+++ b/framework/src/binding/sdk/mod.rs
@@ -156,7 +156,7 @@ impl<S: 'static + ServiceState, C: ChainQuerier, D: Dispatcher> ServiceSDK
             payload.to_string(),
         );
 
-        let result = self.dispatcher.call(ctx, true)?;
+        let result = self.dispatcher.read(ctx)?;
         if result.is_error {
             Err(SDKError::DispatchFailed { error: result.ret }.into())
         } else {
@@ -180,7 +180,7 @@ impl<S: 'static + ServiceState, C: ChainQuerier, D: Dispatcher> ServiceSDK
             payload.to_string(),
         );
 
-        let result = self.dispatcher.call(ctx, false)?;
+        let result = self.dispatcher.write(ctx)?;
         if result.is_error {
             Err(SDKError::DispatchFailed { error: result.ret }.into())
         } else {
--- a/framework/src/binding/sdk/mod.rs
+++ b/framework/src/binding/sdk/mod.rs
@@ -5,33 +5,39 @@ pub use chain_querier::{ChainQueryError, DefaultChainQuerier};
 use std::cell::RefCell;
 use std::rc::Rc;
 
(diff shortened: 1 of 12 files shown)


framework/src/binding/sdk/mod.rs
```diff
@@ -136,13 +142,62 @@ impl<S: 'static + ServiceState, C: ChainQuerier> ServiceSDK for DefalutServiceSD
...
+        payload: &str,
+    ) -> ProtocolResult<String> {
+        let ctx = ServiceContext::with_context(
+            ctx,
+            service.to_string(),
+            method.to_string(),
+            payload.to_string(),
+        );
+
+        let result = self.dispatcher.call(ctx, true)?;
```
zeroqn: Better replace bool with enum.
For example:
```rust
self.dispatcher.call(ctx, Call::Readonly)?;
```

framework/src/binding/sdk/mod.rs
```diff
@@ -136,13 +142,62 @@ impl<S: 'static + ServiceState, C: ChainQuerier> ServiceSDK for DefalutServiceSD
...
+        payload: &str,
+    ) -> ProtocolResult<String> {
+        let ctx = ServiceContext::with_context(
+            ctx,
+            service.to_string(),
+            method.to_string(),
+            payload.to_string(),
+        );
+
+        let result = self.dispatcher.call(ctx, false)?;
```
KaoImin: Using `bool` type here seems perplexed.

protocol/src/traits/executor.rs
```diff
@@ -52,3 +54,15 @@ pub trait Executor {
...
     ) -> ProtocolResult<ExecResp>;
 }
+
+pub trait Dispatcher {
+    fn call(&self, context: ServiceContext, readonly: bool) -> ProtocolResult<ExecResp>;
+}
+
+pub struct DispatcherHolder;
+
+impl Dispatcher for DispatcherHolder {
```
yejiayu: More reasonable to name it `NoopDispatcher`.

framework/src/binding/sdk/mod.rs
mkxbl: add comments to `Dispatcher` trait

protocol/src/traits/executor.rs
mkxbl: renamed

PR #88, opened 2020-01-03
[ᚬmaster] feat(trace): add moodyblue trace

KaoImin:
**What type of PR is this?**
feature

* add moodyblus trace in consensus module
* let bls private key extended from secp256k1 private key 

Fixes #

### DIFF
--- a/core/consensus/src/synchronization.rs
+++ b/core/consensus/src/synchronization.rs
@@ -98,10 +98,10 @@ where
 
             if id != 1 && current_hash != epoch.header.pre_hash {
                 trace::error(
-                    "prev_hash_diff".to_string(),
+                    "diff_prev_hash".to_string(),
                     Some(json!({
                         "epoch_id": id,
-                        "current_prev_hash": epoch.header.pre_hash.as_hex(),
+                        "current_prev_hash":current_hash.as_hex(),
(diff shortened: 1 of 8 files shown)


yejiayu:
Test passed by 4 nodes?

KaoImin:
> Test passed by 4 nodes?

Yes

core/consensus/src/synchronization.rs
```diff
@@ -82,9 +97,25 @@ where
             self.check_proof(id, proof.clone())?;
 
             if id != 1 && current_hash != epoch.header.pre_hash {
+                trace::error(
+                    "prev_hash_diff".to_string(),
```
homura: The event is best described by an action, recommended to use `diff_prev_epoch_hash`

core/consensus/src/synchronization.rs
```diff
@@ -82,9 +97,25 @@ where
...
+                    Some(json!({
+                        "epoch_id": id,
+                        "current_prev_hash": epoch.header.pre_hash.as_hex(),
```
homura: If this should be `current_hash.as_hex()`

KaoImin: Why? This means the `prev_hash` are different.

PR #93, opened 2020-01-05
[ᚬmaster] refactor(consensus): Use RwLock to share consensus status.

yejiayu:
**What type of PR is this?**
refactor

I don't quite understand why the previous code shared the `CurrentConsensusStatus` using both the channel and the read-write lock, which is too weird.

After refactoring, the`CurrentConsensusStatus` will use the read-write lock sharing uniformly.

Fixes #

### DIFF
--- a/core/consensus/src/status.rs
+++ b/core/consensus/src/status.rs
@@ -1,64 +1,46 @@
 use std::sync::Arc;
 
 use derive_more::Display;
-use futures::channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
-use futures::stream::StreamExt;
-use log::{error, warn};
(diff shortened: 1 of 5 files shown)


core/consensus/src/status.rs
```diff
@@ -1,64 +1,46 @@
...
 
-#[derive(Debug)]
-pub struct StatusPivot {
+#[derive(Clone, Debug)]
+pub struct StatusAgent {
     status: Arc<RwLock<CurrentConsensusStatus>>,
-    msg_rx: UnboundedReceiver<UpdateInfo>,
 }
 
-impl StatusPivot {
```
yejiayu: This is where I am mainly modifying.

PR #103, opened 2020-01-08
feat(binding-macro): service method supports none payload and none response

mkxbl:
**What type of PR is this?**

feature: binding-macro

Add feature: service method supports none payload and none response

```rust
#[read] 
fn read_method(&self, ctx: ServiceContext) -> ProtocolResult<()>;

#[write] 
fn write_method(&self, ctx: ServiceContext) -> ProtocolResult<()>;
```

Fixes #

If service method return `()`, the value return from `ServiceSDK`, would be Ok("")

```rust
pub trait ServiceSDK {
    fn read(
        &self,
        ctx: &ServiceContext,
        service: &str,
        method: &str,
        payload: &str,
    ) -> ProtocolResult<String>;

    fn write(
        &mut self,
        ctx: &ServiceContext,
        service: &str,
        method: &str,
        payload: &str,
    ) -> ProtocolResult<String>;
```

### DIFF
--- a/binding-macro/src/lib.rs
+++ b/binding-macro/src/lib.rs
@@ -20,7 +20,7 @@ use crate::service::gen_service_code;
 /// - Verification
 ///  1. Is it a struct method marked with #[service]?
 ///  2. Is visibility private?
-///  3. Parameter signature contains `&self and ctx:ServiceContext`?
+///  3. Parameter signature contains `&self and ctx: ServiceContext`?
 ///  4. Is the return value `ProtocolResult <T: Deserialize + Serialize>` or `ProtocolResult <()>`?
 ///
 /// # Example:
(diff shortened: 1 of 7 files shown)


binding-macro/src/lib.rs
```diff
@@ -20,9 +20,8 @@ use crate::service::gen_service_code;
 /// - Verification
 ///  1. Is it a struct method marked with #[service]?
 ///  2. Is visibility private?
-///  3. Does function generics constrain `fn f<Context: RequestContext>`?
-///  4. Parameter signature contains `&self and ctx:Context`?
-///  5. Is the return value `ProtocolResult <T: Deserialize + Serialize>`?
+///  3. Parameter signature contains `&self and ctx:ServiceContext`?
```
zeroqn: nitpick, space between ```ctx:``` and ```ServiceContext```.

mkxbl: sharp eyes!

fixed

PR #111, opened 2020-01-14
feat(framework): add tx_hash, nonce to ServiceContext

mkxbl:
**What type of PR is this?**

feat(framework)

add tx_hash, nonce to `ServiceContext`, which can be used by `Service`

Fixes #

```rust
impl ServiceContext {
    pub fn get_tx_hash(&self) -> Option<Hash>;
 
    pub fn get_nonce(&self) -> Option<Hash>;
}
```

### DIFF
--- a/protocol/src/types/service_context.rs
+++ b/protocol/src/types/service_context.rs
@@ -3,11 +3,13 @@ use std::rc::Rc;
 
 use derive_more::{Display, From};
 
-use crate::types::{Address, Event};
+use crate::types::{Address, Event, Hash};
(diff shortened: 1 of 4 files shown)


protocol/src/types/service_context.rs
```diff
@@ -70,6 +78,20 @@ impl ServiceContext {
         }
     }
 
+    pub fn get_tx_hash(&self) -> Hash {
+        self.tx_hash
+            .as_ref()
+            .expect("should not get tx hash in query request")
```
huwenchao: `expect` will panic, should return Result or Option

PR #121, opened 2020-01-16
fix duktape pvm_load_args not return proper string

zeroqn:
**What type of PR is this?**
fix

fix duktape pvm_load_args not return proper string

Fixes #


zeroqn:
wrong branch

PR #131, opened 2020-01-31
refactor: change epoch into block

KaoImin:
**What type of PR is this?**
refactor

change epoch into block
change epoch ID into height

Fixes #


devtools/chain/config-node-1.toml
```diff
@@ -0,0 +1,72 @@
+# chain id
+chain_id = "b6a4d7da21443f5e816e8700eea87610e6d769657d6b8ec73028457bf2ca4036"  # by sha256(Muta)
```
yejiayu: delete

devtools/chain/config-node-2.toml
```diff
@@ -0,0 +1,72 @@
+# chain id
```
yejiayu: delete

devtools/chain/config-bootstrap.toml
```diff
@@ -0,0 +1,60 @@
+# chain id
```
yejiayu: delete

devtools/chain/config-node-3.toml
```diff
@@ -0,0 +1,72 @@
+# chain id
```
yejiayu: delete

PR #148, opened 2020-02-07
feat(riscv): add get storage value size syscall

zeroqn:
**What type of PR is this?**
feat
get storage value size

Fixes #


huwenchao:
Why is this needed? It can be implemented with get_storage, with value pointer passed with NULL.

zeroqn:
> Why is this needed? It can be implemented with get_storage, with value pointer passed with NULL.

Writing bytes to NULL pointer? I don't think it's a good idea.

PR #162, opened 2020-02-11
feat(config): move bls_pub_key config to genesis.toml

mkxbl:
**What type of PR is this?**

feat: adjust bls_pub_key config

move bls_pub_key config to genesis.toml, so we can update validator through service

Fixes #

### DIFF
--- a/devtools/chain/genesis.toml
+++ b/devtools/chain/genesis.toml
@@ -3,8 +3,37 @@ prevhash = "44915be5b6c20b0678cf05fcddbbaa832e25d7e6ac538784cd5c24de00d47472"
 
 [[services]]
 name = "asset"
-payload = '{ "id": "f56924db538e77bb5951eb5ff0d02b88983c49c45eea30e8ae3e7234b311436c", "name": "MutaToken", "symbol": "MT", "supply": 320000011, "issuer": "f8389d774afdad8755ef8e629e5a154fddc6325a" }'
+payload = '''
+{
+    "id": "f56924db538e77bb5951eb5ff0d02b88983c49c45eea30e8ae3e7234b311436c",
+    "name": "MutaToken",
+    "symbol": "MT",
+    "supply": 320000011,
+    "issuer": "f8389d774afdad8755ef8e629e5a154fddc6325a"
+}
+'''
 
 [[services]]
 name = "metadata"
-payload = '{ "chain_id": "b6a4d7da21443f5e816e8700eea87610e6d769657d6b8ec73028457bf2ca4036", "common_ref": "703873635a6b51513451", "timeout_gap": 20, "cycles_limit": 99999999, "cycles_price": 1, "interval": 3000, "verifier_list": [{"address": "f8389d774afdad8755ef8e629e5a154fddc6325a", "propose_weight": 1, "vote_weight": 1}], "propose_ratio": 15, "prevote_ratio": 10, "precommit_ratio": 10, "brake_ratio": 7}'
+payload = '''
+{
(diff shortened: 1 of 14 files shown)


devtools/chain/genesis.toml
```diff
@@ -3,8 +3,37 @@ prevhash = "44915be5b6c20b0678cf05fcddbbaa832e25d7e6ac538784cd5c24de00d47472"
...
+{
+    "chain_id": "b6a4d7da21443f5e816e8700eea87610e6d769657d6b8ec73028457bf2ca4036",
+    "common_ref": "703873635a6b51513451",
+    "timeout_gap": 20,
+    "cycles_limit": 999999999999,
+    "cycles_price": 1,
+    "interval": 3000,
+    "verifier_list": [
+        {
+            "pub_key": "04188ef9488c19458a963cc57b567adde7db8f8b6bec392d5cb7b67b0abc1ed6cd966edc451f6ac2ef38079460eb965e890d1f576e4039a20467820237cda753f07a8b8febae1ec052190973a1bcf00690ea8fc0168b3fbbccd1c4e402eda5ef22",
```
KaoImin: Can writing public keys in the genesis block guaranteed the order?

mkxbl: It's a fixed order. Need a ascending order ?

KaoImin: The public keys' order needs to be consistent with the validator list.

huwenchao: The whole `verifier` struct contains both pub_key and address, so the order should always be consistent.
By the way, I recommend rename pub_key to bls_pub_key to prevent misunderstanding.

KaoImin: > The whole `verifier` struct contains both pub_key and address, so the order should always be consistent.
> By the way, I recommend rename pub_key to bls_pub_key to prevent misunderstanding.

A good idea.

src/default_start.rs
```diff
@@ -248,14 +268,11 @@ pub async fn start<Mapping: 'static + ServiceMapping>(
...
     let mut bls_pub_keys = HashMap::new();
-    for (validator, bls_pub_key) in verifier_list
-        .iter()
-        .zip(config.consensus.public_keys.iter())
-    {
-        let address = validator.address.as_bytes();
-        let hex_pubkey = hex::decode(bls_pub_key).map_err(MainError::FromHex)?;
+    for validator_extend in metadata.verifier_list.iter() {
+        let address = validator_extend.address.as_bytes();
+        let hex_pubkey =
```
huwenchao: It only changes how bls_pub_key is read when starting. Should check that when bls_pub_key changes in metadata service, the consensus module gets the latest bls_pub_key. 
You can refer how consensus module gets the latest verifier list.

protocol/src/types/primitive.rs
```diff
@@ -232,13 +232,21 @@ pub struct Metadata {
...
-    pub verifier_list:   Vec<Validator>,
+    pub verifier_list:   Vec<ValidatorExtend>,
     pub propose_ratio:   u64,
     pub prevote_ratio:   u64,
     pub precommit_ratio: u64,
     pub brake_ratio:     u64,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct ValidatorExtend {
```
huwenchao: Why use a new struct here other than add the `bls_pub_key` field into the existing `Validator` struct?

devtools/chain/genesis.toml
mkxbl: rename pub_key to bls_pub_key

PR #169, opened 2020-02-12
feat(sync): Trigger sync after waiting for consensus interval

yejiayu:
**What type of PR is this?**
feat

If the node is only 1 height behind, wait for a period of time (equals consensus interval) first, and enter synchronization if it is still 1 height behind.

Fixes #


core/consensus/src/synchronization.rs
```diff
@@ -351,4 +344,36 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
...
+            return Ok(false);
+        }
+
+        if remote_height <= current_height {
+            return Ok(false);
+        }
+
+        if current_height == remote_height - 1 {
+            let status = self.status.to_inner();
+            Delay::new(Duration::from_millis(status.consensus_interval)).await;
```
KaoImin: The delay duration seems a little short.

PR #185, opened 2020-02-14
fix:(mempool): add info in mempool

rev-chaos:
**What type of PR is this?**
fix

Fixes #


core/mempool/src/tx_cache.rs
```diff
@@ -303,6 +303,7 @@ impl TxCache {
 
         // If queue inserts into queue failed, removes from map.
         if rst.is_err() {
+            log::error!("{:?} insert queue failed! {:?}", &tx_hash, rst);
```
yejiayu: add prefix [core_mempool]

PR #189, opened 2020-02-19
fix(mempool): fix repeat txs, add flush_incumbent_queue

rev-chaos:
**What type of PR is this?**
fix

1. add a RwLock to avoid inserting tx during flush_memory
2. flush incumbent queue in each flush operation
3. fix insert propose tx

Fixes #

### DIFF
--- a/core/mempool/src/lib.rs
+++ b/core/mempool/src/lib.rs
@@ -108,7 +108,7 @@ where
     async fn package(&self, ctx: Context, cycle_limit: u64) -> ProtocolResult<MixedTxHashes> {
         let current_height = self.adapter.get_latest_height(ctx.clone()).await?;
         log::info!(
-            "[mempool]: {:?} txs in map and {:?} txs in queue while package",
+            "[core_mempool]: {:?} txs in map and {:?} txs in queue while package",
             self.tx_cache.len(),
             self.tx_cache.queue_len(),
         );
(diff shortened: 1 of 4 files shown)


core/mempool/src/lib.rs
```diff
@@ -40,6 +41,8 @@ pub struct HashMemPool<Adapter: MemPoolAdapter> {
     callback_cache: Map<SignedTransaction>,
     /// Supply necessary functions from outer modules.
     adapter:        Adapter,
+    /// exclusive flush_memory and insert_tx to avoid repeat txs insertion.
+    flush_lock:     RwLock<bool>,
```
KaoImin:
```suggestion
    flush_lock:     RwLock<()>,
```

core/mempool/src/lib.rs
```diff
@@ -53,6 +56,7 @@ where
             tx_cache: TxCache::new(pool_size * 2),
             callback_cache: Map::new(pool_size),
             adapter,
+            flush_lock: RwLock::new(true),
```
KaoImin:
```suggestion
            flush_lock: RwLock::new(),
```

core/mempool/src/tx_cache.rs
```diff
@@ -203,12 +204,16 @@ impl TxCache {
...
                 }
                 // After previous filter, tx are valid and should cache in temp_queue.
-                queue_role
+                if let Err(e) = queue_role
                     .candidate
                     .push(Arc::<TxWrapper>::clone(&shared_tx))
-                    .map_err(|_| MemPoolError::InsertCandidate {
-                        len: queue_role.candidate.len(),
-                    })?;
+                {
```
yejiayu: If the queue is already full, what is the point of continuing loop?

rev-chaos: When this happens, there are only a few transactions remaining, so even if the loop goes down, it will basically not consume performance. However, if you use the jump-out method, you need a more complicated design, which is not necessary.

rev-chaos: But I should remove these pop-out txs from the map to avoid a memory leak.

PR #194, opened 2020-02-21
feat(api): Supports configuring the max number of connections.

yejiayu:
**What type of PR is this?**
feat

- **Replace tide with actix-web**
Because tide incorrectly handled the `too many file opens` error, when this error occurred, tide directly ended the` loop {income.accept} `, causing the entire server to exit.

- **Maximum connection**
Added a maximum connection configuration to prevent possible DDOS attacks.

Fixes #


src/default_start.rs
```diff
@@ -423,21 +427,18 @@ pub async fn start<Mapping: 'static + ServiceMapping>(
         graphql_config.maxconn = config.graphql.maxconn;
     }
 
-    // Run execute demon
-    tokio::spawn(async move {
-        futures::executor::block_on(exec_demon.run());
+    let actix_handle = std::thread::spawn(move || {
```
KaoImin: The `graphql` process will hold on a thread consistently.

core/api/src/config.rs
```diff
@@ -6,6 +6,15 @@ pub struct GraphQLConfig {
...
     pub graphiql_uri: String,
+
+    // Set number of workers to start.
+    // By default http server uses number of available logical cpu as threads count.
+    pub workers: usize,
+
+    // Sets the maximum per-worker number of concurrent connections.
+    // All socket listeners will stop accepting connections when this limit is reached for each
+    // worker. By default max connections is set to a 25k.
+    pub maxconn: usize,
```
mohanson: A socket will occupy a file descriptor, so it will failed at a unadjusted system which the default fd count is 1024.

yejiayu: The default of 1024 is not sufficient and will need to be adjusted to a larger value in most scenarios.

src/default_start.rs
```diff
@@ -421,7 +411,34 @@ pub async fn start<Mapping: 'static + ServiceMapping>(
...
+
+    let actix_handle = std::thread::spawn(move || {
+        // Run GraphQL server
+        actix_rt::System::new("muta-graphql").block_on(async move {
+            core_api::start_graphql(graphql_config, api_adapter).await;
+        });
+    });
+
+    match actix_handle.join() {
+        Ok(()) => std::process::exit(0),
```
mohanson: I suspect that a data error will occur on restart

mohanson: I suspect a data consistency error will occur when restart

core/api/src/config.rs
```diff
@@ -17,6 +26,8 @@ impl Default for GraphQLConfig {
 
             graphql_uri:  "/graphql".to_owned(),
             graphiql_uri: "/graphiql".to_owned(),
+            workers:      num_cpus::get(),
+            maxconn:      25000,
```
mohanson: the default maxconn could be less because the default fd count on Linux is 1024

src/default_start.rs
yejiayu: why?

PR #195, opened 2020-02-21
feat: check size and cycle limit when insert tx into mempool

huwenchao:
**What type of PR is this?**
New feature.

- Add tx size check to prevent attacking by composing big tx
- Add tx_num_limit to genesis, and it will pack at most this number of tx into a block.
- Change the meaning of cycles_limit in genesis. It becomes the limit of every tx instead of the whole block.

Fixes #

### DIFF
--- a/core/mempool/src/tx_cache.rs
+++ b/core/mempool/src/tx_cache.rs
@@ -225,7 +225,7 @@ impl TxCache {
                 tx_count += 1;
                 if tx_count > tx_num_limit {
                     stage = stage.next();
-                    tx_count = 0;
+                    tx_count = 1;
                 }
 
                 match stage {
--- a/core/mempool/src/tx_cache.rs
+++ b/core/mempool/src/tx_cache.rs
@@ -178,6 +178,7 @@ impl TxCache {
 
     pub fn package(
         &self,
+        _cycles_limit: u64,
         tx_num_limit: u64,
         current_height: u64,
         timeout: u64,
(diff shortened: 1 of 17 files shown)


protocol/src/traits/consensus.rs
```diff
@@ -114,8 +114,7 @@ pub trait CommonConsensusAdapter: Send + Sync {
         timestamp: u64,
     ) -> ProtocolResult<Metadata>;
 
-    /// Set timeout_gap in mempool.
-    fn set_timeout_gap(&self, context: Context, timeout_gap: u64);
+    fn set_args(&self, _context: Context, timeout_gap: u64, cycles_limit: u64, max_tx_size: u64);
```
KaoImin:
```suggestion
    fn set_args(&self, context: Context, timeout_gap: u64, cycles_limit: u64, max_tx_size: u64);
```

core/mempool/src/tx_cache.rs
```diff
@@ -215,12 +215,10 @@ impl TxCache {
...
                 }
-                // Accumulate cycles. The order_tx_hashes and the propose_tx_hashes both collect
-                // transactions under cycle limit.
-                cycle_count += shared_tx.tx.raw.cycles_limit;
-                if cycle_count > cycle_limit {
+                tx_count += 1;
+                if tx_count > tx_num_limit {
                     stage = stage.next();
-                    cycle_count = shared_tx.tx.raw.cycles_limit;
+                    tx_count = 0;
```
rev-chaos:
```suggestion
                    tx_count = 1;
```

rev-chaos:
Lack of setting tx_num_limit and max_tx_size process, and the package interface of mempool should also be modified

huwenchao:
> Lack of setting tx_num_limit and max_tx_size process, and the package interface of mempool should also be modified

`set_timeout_gap` was changed to  `set_args`, which includes setting tx_num_limit and max_tx_size process.

PR #196, opened 2020-02-22
refactor(network): peer manager

zeroqn:
**What type of PR is this?**
refactor

- more cleaner peer state management
- enforce peer id in multiaddr
- enforce max connections
- allow protected peers' incoming connections regardless max connections
- a lot of unit tests
- skip sending data to blocked sessions
- whitelist peers suppor
- whitelist peers only
.. etc

Fixes #

The major logic reside in peer_manager/mod.rs, we can start from here.
test_manager.rs can be used as reference to verify logic.


yejiayu:
The chain failed after running for an hour. @zeroqn

core/network/src/peer_manager/mod.rs
```diff
@@ -36,403 +48,482 @@ use log::{debug, error, info, warn};
...
+        self.peers.write().insert(peer.clone());
+        if let Some(chain_addr) = peer.owned_chain_addr() {
+            self.chain.write().insert(chain_addr, peer);
         }
     }
 
-    pub fn disconnect_peer(&self, peer_id: &PeerId) {
-        self.connecting.write().remove(peer_id);
-        self.connected.write().remove(peer_id);
+    pub(self) fn restore(&self, peers: Vec<ArcPeer>) {
```
yejiayu: Rust has a lot of visibility, such as pub(crate), pub(self), and so on. I think this visibility adds an extra burden when review code, and most of the time pub and private are perfectly adequate.

What do you think?

zeroqn:
> The chain failed after running for an hour. @zeroqn

Cpu outage, will rerun it.

PR #218, opened 2020-03-10
feat: add tx hook in framework

huwenchao:
**What type of PR is this?**
New feature.

Add tx hook in framework, can be used to implement fee logic.

Fixes #


framework/src/executor/mod.rs
```diff
@@ -234,9 +234,25 @@ impl<S: 'static + Storage, DB: 'static + TrieDB, Mapping: 'static + ServiceMappi
...
+            let sdk = self.get_sdk(&name)?;
+            let tx_hook_service = self.service_mapping.get_service(name.as_str(), sdk)?;
+            tx_hook_services.push(tx_hook_service);
+        }
+
         let result = match exec_type {
             ExecType::Read => service.read_(context),
-            ExecType::Write => service.write_(context),
+            ExecType::Write => {
+                for tx_hook_service in tx_hook_services.iter_mut() {
```
mkxbl: I doubt this is not the right place for executor to exec tx_hook logic. Tx_hook( before or after) logic should be executed only one time for a tx lifetime. But if logic code wrote in here,  it will be executed every time when a service call another service , because `fn sdk.write` will trigger  it.

yejiayu: Insert the `tx_hook` here. https://github.com/nervosnetwork/muta/blob/fda4204f52b4214078b1ef5894b950b1937ea778/framework/src/executor/mod.rs#L290

yejiayu:
Add some unit tests to `asset_service` to make sure the logic is correct

huwenchao:
> Add some unit tests to `asset_service` to make sure the logic is correct

They are in the huobi-chain repo: <https://github.com/HuobiGroup/huobi-chain/pull/43/files#diff-a272043483840b9b5a3696a674171a9aR113>. `asset_service` in muta does not have this function.

Also, I added some unit tests for the hook function.

framework/src/executor/mod.rs
```diff
@@ -227,6 +231,27 @@ impl<S: 'static + Storage, DB: 'static + TrieDB, Mapping: 'static + ServiceMappi
...
+        let mut tx_hook_services = vec![];
+        for name in self.service_mapping.list_service_name().into_iter() {
+            let sdk = self.get_sdk(&name)?;
+            let tx_hook_service = self.service_mapping.get_service(name.as_str(), sdk)?;
+            tx_hook_services.push(tx_hook_service);
+        }
+        for tx_hook_service in tx_hook_services.iter_mut() {
+            tx_hook_service.tx_hook_before_(context.clone())?;
+        }
+        let original_res = self.call(context.clone(), exec_type);
```
yejiayu: `tx_hook_after` might make different choices depending on whether or not an error occurred in the execution, but for now we can ignore it.

Adding a TODO ensures that it will not be forgotten in the future when it is perfected.

PR #221, opened 2020-03-11
feat(sync): Split a transaction in a block into multiple requests

yejiayu:
**What type of PR is this?**
feat

In a weak network environment, a single request for a large number of transactions causes the request to timeout

Splitting a large number of transactions into several requests can greatly alleviate this situation

Fixes #

### DIFF
--- a/core/consensus/src/synchronization.rs
+++ b/core/consensus/src/synchronization.rs
@@ -29,6 +29,8 @@ pub struct OverlordSynchronization<Adapter: SynchronizationAdapter> {
     status:  StatusAgent,
     lock:    Arc<Mutex<()>>,
     syncing: Mutex<()>,
+
+    sync_txs_chunk_size: usize,
 }
 
 #[async_trait]
@@ -102,7 +104,12 @@ impl<Adapter: SynchronizationAdapter> Synchronization for OverlordSynchronizatio
 }
 
 impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
-    pub fn new(adapter: Arc<Adapter>, status: StatusAgent, lock: Arc<Mutex<()>>) -> Self {
+    pub fn new(
+        sync_txs_chunk_size: usize,
+        adapter: Arc<Adapter>,
+        status: StatusAgent,
+        lock: Arc<Mutex<()>>,
+    ) -> Self {
         let syncing = Mutex::new(());
 
         Self {
@@ -110,6 +117,8 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
             status,
             lock,
             syncing,
+
+            sync_txs_chunk_size,
         }
     }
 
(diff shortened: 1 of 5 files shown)


core/consensus/src/synchronization.rs
```diff
@@ -230,10 +239,18 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
...
+        let mut txs = Vec::with_capacity(block.ordered_tx_hashes.len());
+
+        for tx_hashes in block.ordered_tx_hashes.chunks(self.sync_txs_chunk_size) {
+            let remote_txs = self
+                .adapter
+                .get_txs_from_remote(ctx.clone(), &tx_hashes)
+                .await?;
+            for tx in remote_txs.into_iter() {
+                txs.push(tx);
+            }
```
huwenchao: Consider `extend`?

yejiayu: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice
> Clones and appends all elements in a slice to the Vec.

`clone` here is redundant.

huwenchao: I mean this:
https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend

```rust
txs.extend(remote_txs);
```

`remote_txs` is moved after this, so I think it should not be cloned.

ref: https://stackoverflow.com/questions/40792801/best-way-to-concatenate-vectors-in-rust

huwenchao: I wrote a benchmark, the performance of 3 ways differs little.

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};

const vec_len: usize = 5000;
const str_len: usize = 1024;

fn extend(mut a: Vec<String>, mut b: Vec<String>) {
    a.extend(b);
}

fn append(mut a: Vec<String>, mut b: Vec<String>) {
    a.append(&mut b);
}

fn push(mut a: Vec<String>, mut b: Vec<String>) {
    for i in b.into_iter() {
        a.push(i);
    }
}

fn bench(c: &mut Criterion) {
    let m: Vec<String> = Vec::with_capacity(vec_len);
    let n = vec!["a".repeat(str_len); vec_len];
    c.bench_function("extend", |b| {
        b.iter(|| extend(black_box(m.clone()), black_box(n.clone())));
    });
    c.bench_function("append", |b| {
        b.iter(|| append(black_box(m.clone()), black_box(n.clone())));
    });
    c.bench_function("push", |b| {
        b.iter(|| push(black_box(m.clone()), black_box(n.clone())));
    });
}

criterion_group!(benches, bench);
criterion_main!(benches);
```

Result:
```
extend                  time:   [2.9711 ms 3.0033 ms 3.0381 ms]
                        change: [-93.354% -93.291% -93.221%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe

append                  time:   [2.8076 ms 2.8323 ms 2.8608 ms]
                        change: [-93.626% -93.566% -93.494%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe

push                    time:   [2.8295 ms 2.8521 ms 2.8767 ms]
                        change: [-93.713% -93.655% -93.594%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```

So it is just a choice of personal taste.

yejiayu: `extend` looks more concise.

PR #222, opened 2020-03-12
feat: re-execute blocks to recover current status

KaoImin:
**What type of PR is this?**
feature

* Re-execute blocks from the latest exec_height to current_height to recover current status

Fixes #

### DIFF
--- a/core/consensus/src/wal.rs
+++ b/core/consensus/src/wal.rs
@@ -65,7 +65,7 @@ impl SignedTxsWAL {
         let _ = file
             .read_to_end(&mut read_buf)
             .map_err(ConsensusError::WALErr)?;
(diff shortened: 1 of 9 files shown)


core/consensus/src/wal.rs
```diff
@@ -84,7 +84,7 @@ impl SignedTxsWAL {
                 ConsensusError::Other(format!("parse folder name {:?} error {:?}", folder, err))
             })?;
 
-            if height < till {
+            if height <= till {
```
yejiayu: I don't understand the variable name `till` very well.

KaoImin: Till means up to the time. In this situation, means remove the transactions of the height that lower than `till`.

yejiayu: Don't use this ambiguous name and consider `committed_height`.

PR #228, opened 2020-03-18
fix(storage): fix rocksdb too many open files error

mohanson:
**What type of PR is this?**

Bug fix

Add an option to set rocksdb's max_open_files. 

By default, rocksdb's max_open_files = -1, which means  "keep open all files".

And set rocksdb's target_file_size_base == 1G.

Fixes #


yejiayu:
from rocksdb doc
> target_file_size_base and target_file_size_multiplier -- Files in level 1 will have target_file_size_base bytes. Each next level's file size will be target_file_size_multiplier bigger than previous one. However, by default target_file_size_multiplier is 1, so files in all L1..Lmax levels are equal. Increasing target_file_size_base will reduce total number of database files, which is generally a good thing. We recommend setting target_file_size_base to be max_bytes_for_level_base / 10, so that there are 10 files in level 1.

`target_file_size_base` looks like it needs to match `max_bytes_for_level_base`

mohanson:
Yes, a mistake. We could just use the default values for `target_file_size`.

lgn21st:
@mohanson 从代码上看，把 rocksdb 的 default_max_open_files 设置为 64，那么实际在执行的时候，这个数值应该是多少比较合适呢？有办法通过 benchmark 来 proof 一个合理的数值么？

PR #229, opened 2020-03-18
fix(config): use serde(default) for rocksdb conf

mohanson:
**What type of PR is this?**

bug fix

image

Set default value for ConfigRocksDB, so that we don't need to add a new section "rocksdb" in configuration files.

Fixes #


yejiayu:
Add the rocksdb configuration to `config.toml`, so the user knows that there's a configuration item here.

PR #230, opened 2020-03-19
refactor: modify service error handle mechanism

mkxbl:
**What type of PR is this?**

refactor

refactor error handle mechanism to distinguish inner chain error from service error

Fixes #

- Had not used `Bytes` to replace `String` after thinking twice
Payload and response are still `String`, which is easier to read. 
This can be discussed.
- Service call return `ServiceResponse<T>` type
```rust
#[derive(Debug, Clone, Default)]
pub struct ServiceResponse<T: Default> {
    pub code:  u64,
    pub succeed_data:  T,
    pub error_message: String,
}
```
- Breaking change for api
```rust
#[derive(juniper::GraphQLObject, Clone)]
pub struct ServiceResponse {
    pub code:  Uint64,
    pub succeed_data:  String,
    pub error_message: String,
}
```
- Pay attention to the catch panic logic of executor
help me check it and make it right
- TODO: the return type of service hook is not token into account in this PR, need to be refactored next step

### DIFF
--- a/binding-macro/src/service.rs
+++ b/binding-macro/src/service.rs
@@ -113,10 +113,13 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
 
                 match method {
                     #(#list_read_name => {
-                        let payload: #list_read_payload = serde_json::from_str(ctx.get_payload())
-                                .unwrap_or_else(|e| panic!("service macro decode payload failed: {:?}", e));
+                        let payload_res: Result<#list_read_payload, _> = serde_json::from_str(ctx.get_payload());
+                        if let Err(e) = payload_res {
+                            return ServiceResponse::<String>::from_error(1, format!("service macro decode payload failed: {:?}", e));
+                        };
+                        let payload = payload_res.unwrap();
                         let res = self.#list_read_ident(ctx, payload);
-                        if res.code == 0 {
+                        if !res.is_error() {
                             let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
                             if data_json == "null" {
                                 data_json = "".to_owned();
@@ -128,7 +131,7 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
                     },)*
                     #(#list_read_name_nonepayload => {
                         let res = self.#list_read_ident_nonepayload(ctx);
-                        if res.code == 0 {
+                        if !res.is_error() {
                             let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
                             if data_json == "null" {
                                 data_json = "".to_owned();
@@ -148,10 +151,13 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
 
                 match method {
                     #(#list_write_name => {
-                        let payload: #list_write_payload = serde_json::from_str(ctx.get_payload())
-                                .unwrap_or_else(|e| panic!("service macro decode payload failed: {:?}", e));
+                        let payload_res: Result<#list_write_payload, _> = serde_json::from_str(ctx.get_payload());
+                        if let Err(e) = payload_res {
+                            return ServiceResponse::<String>::from_error(1, format!("service macro decode payload failed: {:?}", e));
+                        };
+                        let payload = payload_res.unwrap();
                         let res = self.#list_write_ident(ctx, payload);
-                        if res.code == 0 {
+                        if !res.is_error() {
                             let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
                             if data_json == "null" {
                                 data_json = "".to_owned();
@@ -163,7 +169,7 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
                     },)*
                     #(#list_write_name_nonepayload => {
                         let res = self.#list_write_ident_nonepayload(ctx);
-                        if res.code == 0 {
+                        if !res.is_error() {
                             let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
                             if data_json == "null" {
                                 data_json = "".to_owned();
--- a/binding-macro/src/service.rs
+++ b/binding-macro/src/service.rs
@@ -117,8 +117,10 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
                                 .unwrap_or_else(|e| panic!("service macro decode payload failed: {:?}", e));
                         let res = self.#list_read_ident(ctx, payload);
                         if res.code == 0 {
-                            let data_json = serde_json::to_string(&res.data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
-
+                            let mut data_json = serde_json::to_string(&res.data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
+                            if data_json == "null" {
+                                data_json = "".to_owned();
+                            }
                             ServiceResponse::<String>::from_data(data_json)
                         } else {
                             ServiceResponse::<String>::from_error(res.code, res.error.clone())
@@ -127,8 +129,10 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
                     #(#list_read_name_nonepayload => {
                         let res = self.#list_read_ident_nonepayload(ctx);
                         if res.code == 0 {
-                            let data_json = serde_json::to_string(&res.data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
-
+                            let mut data_json = serde_json::to_string(&res.data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
+                            if data_json == "null" {
+                                data_json = "".to_owned();
+                            }
                             ServiceResponse::<String>::from_data(data_json)
                         } else {
                             ServiceResponse::<String>::from_error(res.code, res.error.clone())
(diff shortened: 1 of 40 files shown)


protocol/src/traits/executor.rs
```diff
@@ -22,12 +22,39 @@ pub struct ExecutorParams {
     pub cycles_limit: u64,
 }
 
-#[derive(Debug, Clone)]
-pub struct ExecResp {
-    pub ret:      String,
-    pub is_error: bool,
+#[derive(Debug, Clone, Default)]
+pub struct ServiceResponse<T: Default> {
```
yejiayu: Use enum?

```
pub enum ServiceResponse {
  Success(T),
  Error({ code: u32, err: String})
}

impl ServiceResponse {
  fn is_err() -> bool
}
```

mkxbl: adjust ServiceResponse to : 
```rust

#[derive(Debug, Clone, Default)]
pub struct ServiceResponse<T: Default> {
    pub code:          u64,
    pub succeed_data:  T,
    pub error_message: String,
}

impl<T: Default> ServiceResponse<T> {
    pub fn from_error(code: u64, error_message: String) -> Self {
        Self {
            code,
            succeed_data: T::default(),
            error_message,
        }
    }

    pub fn from_succeed(succeed_data: T) -> Self {
        Self {
            code: 0,
            succeed_data,
            error_message: "".to_owned(),
        }
    }
}
```

Please check again

binding-macro/src/service.rs
```diff
@@ -113,10 +113,13 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
-                                .unwrap_or_else(|e| panic!("service macro decode payload failed: {:?}", e));
+                        let payload_res: Result<#list_read_payload, _> = serde_json::from_str(ctx.get_payload());
+                        if let Err(e) = payload_res {
+                            return ServiceResponse::<String>::from_error(1, format!("service macro decode payload failed: {:?}", e));
+                        };
+                        let payload = payload_res.unwrap();
                         let res = self.#list_read_ident(ctx, payload);
-                        if res.code == 0 {
+                        if !res.is_error() {
                             let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
```
yejiayu: Returns a fixed error to avoid inconsistencies caused by the `serde`.

mkxbl: fixed

protocol/src/types/service_context.rs
```diff
@@ -96,12 +96,11 @@ impl ServiceContext {
...
     }
 
-    pub fn sub_cycles(&self, cycles: u64) -> ProtocolResult<()> {
+    pub fn sub_cycles(&self, cycles: u64) {
         if self.get_cycles_used() + cycles <= self.cycles_limit {
             *self.cycles_used.borrow_mut() = self.get_cycles_used() + cycles;
-            Ok(())
         } else {
-            Err(ServiceContextError::OutOfCycles.into())
+            panic!("out of cycles");
```
huwenchao: This should be considered as a failed tx instead of a system error.

protocol/src/traits/executor.rs
huwenchao: I think `succeed` and `error_message` are somehow duplicated.
If we want Backward Compatibility, we can still use `ret` and `is_error`, or we can use `code` and `message`, when it is 0, the message is succeed_data, or it is error_message.

huwenchao: > Use enum?
> 
> ```
> pub enum ServiceResponse {
>   Success{ data: T},
>   Error({ code: u32, err: String})
> }
> 
> impl ServiceResponse {
>   fn is_err() -> bool
> }
> ```

The rust style enum is not common in other languages, I think we should avoid using it in interface layer, or it will introduce many troubles in serde/deserde and other sdk.

yejiayu: > The rust style enum is not common in other languages, I think we should avoid using it in interface layer, or it will introduce many troubles in serde/deserde and other sdk.

So why don't you use `c`?

protocol/src/types/service_context.rs
mkxbl: Fixed. see:
```rust
    TokenStream::from(quote! {
        #func_vis fn #func_name#generics(#inputs) #ret {
            if !#request_ident.sub_cycles(#cycles_value) {
                return ServiceResponse::<_>::from_error(2, "cycles macro consume cycles fialed: out of cycles");
            }
            #body
        }
    })
```

protocol/src/traits/executor.rs
mkxbl: > I think `succeed_data` and `error_message` are somehow duplicated.
> If we want Backward Compatibility, we can still use `ret` and `is_error`, or we can use `code` and `message`, when it is 0, the message is succeed_data, otherwise it is error_message.

Because of `code`, i think we can't make backward compatibility.

binding-macro/src/service.rs
```diff
@@ -87,81 +87,99 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
-                        if res_str.as_str() == "null" {
-                            Ok("".to_owned())
+                        let payload_res: Result<#list_read_payload, _> = serde_json::from_str(ctx.get_payload());
+                        if payload_res.is_err() {
+                            return ServiceResponse::<String>::from_error(1, "service macro decode payload failed".to_owned());
+                        };
+                        let payload = payload_res.unwrap();
+                        let res = self.#list_read_ident(ctx, payload);
+                        if !res.is_error() {
+                            let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
```
yejiayu: Still returns serde error?

binding-macro/src/service.rs
```diff
@@ -87,81 +87,99 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
                     #(#list_read_name => {
-                        let payload: #list_read_payload = serde_json::from_str(ctx.get_payload())
-                                .map_err(|e| protocol::traits::BindingMacroError::JsonParse(e))?;
-                        let res = self.#list_read_ident(ctx, payload)?;
-                        let res_str = serde_json::to_string(&res).map_err(|e| protocol::traits::BindingMacroError::JsonParse(e))?;
-                        if res_str.as_str() == "null" {
-                            Ok("".to_owned())
+                        let payload_res: Result<#list_read_payload, _> = serde_json::from_str(ctx.get_payload());
+                        if payload_res.is_err() {
+                            return ServiceResponse::<String>::from_error(1, "service macro decode payload failed".to_owned());
```
yejiayu: Don't call it service macro because this is a runtime error and not macro error.

binding-macro/src/service.rs
```diff
@@ -87,81 +87,99 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
                         }
                     },)*
                     #(#list_read_name_nonepayload => {
-                        let res = self.#list_read_ident_nonepayload(ctx)?;
-                        let res_str = serde_json::to_string(&res).map_err(|e| protocol::traits::BindingMacroError::JsonParse(e))?;
-                        if res_str.as_str() == "null" {
-                            Ok("".to_owned())
+                        let res = self.#list_read_ident_nonepayload(ctx);
+                        if !res.is_error() {
+                            let mut data_json = serde_json::to_string(&res.succeed_data).unwrap_or_else(|e| panic!("service macro encode payload failed: {:?}", e));
```
yejiayu: ditto

binding-macro/src/service.rs
```diff
@@ -87,81 +87,99 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
+                                data_json = "".to_owned();
+                            }
+                            ServiceResponse::<String>::from_succeed(data_json)
                         } else {
-                            Ok(res_str)
+                            ServiceResponse::<String>::from_error(res.code, res.error_message.clone())
                         }
                     },)*
-                    _ => Err(protocol::traits::BindingMacroError::NotFoundMethod{ service: service.to_owned(), method: method.to_owned() }.into())
+                    _ => panic!("service macro not found method:{:?} of service:{:?}", method, service)
```
yejiayu: Should not panic, method not found is normal.

binding-macro/src/service.rs
```diff
@@ -87,81 +87,99 @@ pub fn gen_service_code(_: TokenStream, item: TokenStream) -> TokenStream {
...
```
yejiayu: ditto

binding-macro/src/service.rs
```diff
@@ -275,7 +293,7 @@ fn get_genesis_body(item: &ImplItemMethod) -> proc_macro2::TokenStream {
 
                 quote!{
                     let payload: #payload_ident = serde_json::from_str(&_payload)
-                                    .map_err(|e| protocol::traits::BindingMacroError::JsonParse(e))?;
+                    .unwrap_or_else(|e| panic!("service macro decode payload failed: {:?}", e));
```
yejiayu: Panic if serialization fails? 

So, if I randomly pass a wrong string, the chain will crash?

framework/src/executor/mod.rs
```diff
@@ -217,62 +226,86 @@ impl<S: 'static + Storage, DB: 'static + TrieDB, Mapping: 'static + ServiceMappi
...
-            Ok(resp) => resp,
-            Err(e) => ExecResp {
-                ret:      e.to_string(),
-                is_error: true,
-            },
+    fn catch_call(
+        &mut self,
+        context: ServiceContext,
+        exec_type: ExecType,
+    ) -> ProtocolResult<ServiceResponse<String>> {
```
yejiayu: Remove this retry logic, executor  shouldn't care retry.

binding-macro/src/service.rs
mkxbl: This is for `fn init_genesis` only, if init genesis payload is wrong, i think should panic and check( make it right ) genesis config and code

yejiayu: right

binding-macro/src/service.rs
mkxbl: fixed:
```rust
return ServiceResponse::<String>::from_error(1, "decode service payload failed".to_owned());
```

binding-macro/src/service.rs
mkxbl: fixed:
```rust
_ => ServiceResponse::<String>::from_error(2, format!("not found method:{:?} of service:{:?}", method, service))
```

binding-macro/src/service.rs
mkxbl: ditto fixed

framework/src/executor/mod.rs
mkxbl: fixed

PR #239, opened 2020-04-01
feat(consensus): add trust metric for consensus

KaoImin:
**What type of PR is this?**
feature

* add trust metric for consensus

Fixes #


core/consensus/src/engine.rs
```diff
@@ -413,6 +414,17 @@ impl<Adapter: ConsensusAdapter + 'static> Wal for ConsensusEngine<Adapter> {
     }
 }
 
+impl<Adapter: ConsensusAdapter + 'static> Report for ConsensusEngine<Adapter> {
+    fn report_error(&self, ctx: Context, err: OverlordError) {
+        match err {
+            OverlordError::CryptoErr(_) | OverlordError::AggregatedSignatureErr(_) => self
+                .adapter
+                .report_bad(ctx, TrustFeedback::Bad(err.to_string())),
```
zeroqn, line 422: Bad? Not worse?

core/consensus/src/message.rs
```diff
@@ -85,10 +85,13 @@ impl<C: Consensus + 'static> ProposalMessageHandler<C> {
 impl<C: Consensus + 'static> MessageHandler for ProposalMessageHandler<C> {
     type Message = Proposal;
 
-    async fn process(&self, ctx: Context, msg: Self::Message) {
+    async fn process(&self, ctx: Context, msg: Self::Message) -> TrustFeedback {
         if let Err(e) = self.consensus.set_proposal(ctx, msg.0).await {
```
zeroqn, line 89: Should we match err here? exclude consensus internal error?

zeroqn, line 89: Does this exclude consensus internal error?

core/consensus/src/message.rs
```diff
@@ -85,10 +85,13 @@ impl<C: Consensus + 'static> ProposalMessageHandler<C> {
...
             warn!("set proposal {}", e);
+            return TrustFeedback::Bad(e.to_string());
         }
+
+        TrustFeedback::Neutral
```
zeroqn, line 94: IMO, this should report TrustFeedback::Good.

core/consensus/src/message.rs
KaoImin: Yes, the error can be only caused by `DecodeErr`.

zeroqn:
Close, will make another pr.

PR #240, opened 2020-04-02
refactor(protocol): impl fixed codec trait by proc macro

KaoImin:
**What type of PR is this?**
refactor

* impl fixed codec trait by proc macro

Fixes #


protocol/Cargo.toml
```diff
@@ -28,6 +28,7 @@ rlp = "0.4"
 cita_trie = "2.0"
 json = "0.12"
 byteorder = "1.3"
+fixed-codec-derive = { git = "https://github.com/nervosnetwork/muta-utils.git", branch = "master" }
```
yejiayu, line 31: publish to crates.io?

KaoImin:
> If i write:
> 
> ```rust
> #[derive(RlpFixedCodec)]
> pub struct Test {
>     pub a: String,
>     pub b: Kind
> }
> 
> #[derive(RlpFixedCodec)]
> pub enum Kind {
>     Good,
>     Bad
> }
> ```
> 
> Could this work ?

This can not work. The `RlpFixedCodec` macro only support `struct` type.

protocol/Cargo.toml
KaoImin: Already done.

protocol/src/fixed_codec/primitive.rs
```diff
@@ -65,6 +78,32 @@ impl FixedCodec for u64 {
...
+
+        Ok(BytesMut::from(buf.as_ref()).freeze())
+    }
+
+    fn decode_fixed(bytes: Bytes) -> ProtocolResult<Self> {
+        Ok(LittleEndian::read_u128(bytes.as_ref()))
+    }
+}
+
+impl FixedCodec for usize {
```
yejiayu, line 94: The "usize"  length differently on 32-bit and 64-bit systems, so it is not recommended.

Delete this code.

KaoImin: I have already fixed the length of `usize` as 64 bytes.

KaoImin:
ping @yejiayu

PR #251, opened 2020-04-21
fix(binding):overflow of StoreUint64 trait, modify foo_ to inner_foo

fendouhyz:
**What type of PR is this?**
bug
fix bug of StoreUint64 overflow panic

Fixes #

### DIFF
--- a/framework/src/binding/store/primitive.rs
+++ b/framework/src/binding/store/primitive.rs
@@ -7,8 +7,6 @@ use protocol::traits::{ServiceState, StoreBool, StoreString, StoreUint64};
 use protocol::types::Hash;
 use protocol::ProtocolResult;
 
-use crate::binding::store::StoreError;
-
 pub struct DefaultStoreBool<S: ServiceState> {
     state: Rc<RefCell<S>>,
     key:   Hash,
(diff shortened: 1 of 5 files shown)


framework/src/binding/store/primitive.rs
```diff
@@ -77,132 +77,138 @@ impl<S: ServiceState> DefaultStoreUint64<S> {
...
     }
 
     // Add val with self
     // And set the result back to self
-    fn add(&mut self, val: u64) {
-        self.add_(val)
-            .unwrap_or_else(|e| panic!("StoreUint64 add value failed: {}", e));
+    fn add(&mut self, val: u64) -> bool {
+        let mut overflow = false;
+        self.inner_add(val).unwrap_or_else(|_| overflow = true);
```
yejiayu, line 170: Consider the case where an error occurred other than an overflow, such as IO.

framework/src/binding/store/primitive.rs
```diff
@@ -77,132 +77,138 @@ impl<S: ServiceState> DefaultStoreUint64<S> {
...
     // And set the result back to self
-    fn mul_(&mut self, val: u64) -> ProtocolResult<()> {
-        let sv = self.get_()?;
+    fn inner_mul(&mut self, val: u64) -> ProtocolResult<()> {
+        let sv = self.inner_get()?;
 
         match val.overflowing_mul(sv) {
-            (mul, false) => self.set_(mul),
+            (mul, false) => self.inner_set(mul),
             _ => Err(StoreError::Overflow.into()),
```
yejiayu, line 115: I think this should return true instead of an error

fendouhyz: fixed

framework/src/binding/store/primitive.rs
fendouhyz: fixed

framework/src/binding/store/primitive.rs
```diff
@@ -65,144 +63,157 @@ impl<S: ServiceState> DefaultStoreUint64<S> {
...
+            false
         }
     }
 }
 
 impl<S: ServiceState> StoreUint64 for DefaultStoreUint64<S> {
     fn get(&self) -> u64 {
-        self.get_()
-            .unwrap_or_else(|e| panic!("StoreUint64 get failed: {}", e))
+        self.inner_get()
```
mkxbl, line 176: I think we can remove `fn inner_xx` method, and move the logic to `fn xx` trait method.
The reason of `inner_xx` existed is historical.

protocol/src/traits/binding.rs
```diff
@@ -234,27 +234,27 @@ pub trait StoreUint64 {
 
     // Add val with self
     // And set the result back to self
-    fn add(&mut self, val: u64);
+    fn add(&mut self, val: u64) -> bool;
```
mkxbl, line 237: How about name the method `fn safe_xx` ?
Because we had took care of math overflow.

fendouhyz: i think right

framework/src/binding/store/primitive.rs
yejiayu: Agreed, but refactoring may not be included here because the focus of this PR is to solve the overflow.

PR #258, opened 2020-04-27
feat(service):add account service for multi sig

cosinlink:
**What type of PR is this?**
feature
- generate_account, generate an account of multisig which need sigs to be verfied
- get_account_from_address
- verify_signature,when tx----> mempool , mempool----->verify_signature of account service

Fixes #

```
// the data which is provided by user
#[derive(Deserialize, Serialize, Clone, Debug)]
pub struct GenerateAccountPayload {
    pub accounts:  Vec<PayloadAccount>,
    pub threshold: u8,
}


// the data which is stored in stateDB, Account.permission_id and account_type are used for expansion later,you can ignore this 
pub struct Permission {
    pub accounts:  Vec<Account>,
    pub threshold: u8,
}

pub struct Account {
    pub address:       Address,
    pub account_type:  u8,
    pub permission_id: u8,
    pub weight:        u8,
}
```

### DIFF
--- a/built-in-services/account/src/lib.rs
+++ b/built-in-services/account/src/lib.rs
@@ -1,0 +1,146 @@
+use crate::types::{
+    Account, GenerateAccountPayload, GenerateAccountResponse, GetAccountPayload, PayloadAccount,
+    Permission, VerifyPayload, VerifyResponse, ACCOUNT_TYPE_PUBLIC_KEY, MAX_PERMISSION_ACCOUNTS,
+};
+use binding_macro::{cycles, service};
+use bytes::Bytes;
+use hasher::{Hasher, HasherKeccak};
+use protocol::traits::{ExecutorParams, ServiceResponse, ServiceSDK};
+use protocol::types::{Address, Hash, ServiceContext};
+
+#[cfg(test)]
+mod tests;
+pub mod types;
+
+pub struct AccountService<SDK> {
+    sdk: SDK,
+}
+
+#[service]
+impl<SDK: ServiceSDK> AccountService<SDK> {
+    pub fn new(mut sdk: SDK) -> Self {
+        Self { sdk }
+    }
+
+    #[cycles(100_00)]
+    #[read]
+    fn verify(
+        &self,
+        ctx: ServiceContext,
+        payload: VerifyPayload,
+    ) -> ServiceResponse<VerifyResponse> {
+        ServiceResponse::<VerifyResponse>::from_error(
+            110,
+            "accounts length must be [1,16]".to_owned(),
+        )
+    }
+
+    #[cycles(100_00)]
+    #[read]
+    fn get_account_from_address(
+        &self,
+        ctx: ServiceContext,
+        payload: GetAccountPayload,
+    ) -> ServiceResponse<GenerateAccountResponse> {
+        let permission = self
+            .sdk
+            .get_account_value(&payload.user, &0u8)
+            .unwrap_or(Permission {
+                accounts:  Vec::<Account>::new(),
+                threshold: 0,
+            });
+
+        if permission.threshold == 0 {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                110,
+                "account not existed".to_owned(),
+            );
+        }
+
+        let mut accounts = Vec::<PayloadAccount>::new();
+        for item in &permission.accounts {
+            accounts.push(PayloadAccount {
+                address: item.address.clone(),
+                weight:  item.weight,
+            });
+        }
+
+        let response = GenerateAccountResponse {
+            accounts,
+            threshold: permission.threshold,
+            address: payload.user.clone(),
+        };
+
+        ServiceResponse::<GenerateAccountResponse>::from_succeed(response)
+    }
+
+    #[cycles(210_00)]
+    #[write]
+    fn generate_account(
+        &mut self,
+        ctx: ServiceContext,
+        payload: GenerateAccountPayload,
+    ) -> ServiceResponse<GenerateAccountResponse> {
+        if payload.accounts.len() == 0 || payload.accounts.len() > MAX_PERMISSION_ACCOUNTS as usize
+        {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                110,
+                "accounts length must be [1,16]".to_owned(),
+            );
+        }
+
+        let mut weight_all = 0;
+        let mut accounts = Vec::<Account>::new();
+        for item in &payload.accounts {
+            weight_all += item.weight;
+            accounts.push(Account {
+                address:       item.address.clone(),
+                account_type:  ACCOUNT_TYPE_PUBLIC_KEY,
+                permission_id: 0,
+                weight:        item.weight,
+            });
+        }
+
+        if weight_all < payload.threshold || payload.threshold == 0 {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                110,
+                "accounts weight or threshold not valid".to_owned(),
+            );
+        }
+
+        let tx_hash = ctx.get_tx_hash().unwrap();
+
+        let keccak = HasherKeccak::new();
+        let addr_hash = Hash::from_bytes(Bytes::from(keccak.digest(&tx_hash.as_bytes())));
+        if addr_hash.is_err() {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                111,
+                "generate addr_hash from tx_hash failed".to_owned(),
+            );
+        }
+
+        let addr = Address::from_hash(addr_hash.unwrap());
+        if addr.is_err() {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                111,
+                "generate address from tx_hash failed".to_owned(),
+            );
+        }
+
+        let address = addr.unwrap();
+        let permission = Permission {
+            accounts,
+            threshold: payload.threshold,
+        };
+
+        self.sdk.set_account_value(&address, 0u8, permission);
+
+        let response = GenerateAccountResponse {
+            address:   address.clone(),
+            accounts:  payload.accounts,
+            threshold: payload.threshold,
+        };
+
+        ServiceResponse::<GenerateAccountResponse>::from_succeed(response)
+    }
+}
--- a/built-in-services/account/src/lib.rs
+++ b/built-in-services/account/src/lib.rs
@@ -42,7 +42,7 @@ impl<SDK: ServiceSDK> AccountService<SDK> {
         let wit_res: Result<Witness, _> = serde_json::from_str(wit_str.unwrap().as_str());
         if wit_res.is_err() {
             return ServiceResponse::<VerifyResponse>::from_error(
-                112,
+                113,
                 "witness not valid".to_owned(),
             );
         }
@@ -51,7 +51,7 @@ impl<SDK: ServiceSDK> AccountService<SDK> {
         if wit.signature_type == ACCOUNT_TYPE_PUBLIC_KEY {
             if wit.signatures.len() != 1 || wit.pubkeys.len() != 1 {
                 return ServiceResponse::<VerifyResponse>::from_error(
-                    113,
+                    114,
                     "len of signatures or pubkeys must be 1".to_owned(),
                 );
             }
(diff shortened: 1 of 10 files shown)


built-in-services/account/src/types.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Deserialize, Serialize, Clone, Debug)]
+pub struct GenerateAccountPayload {
+    pub accounts:  Vec<PayloadAccount>,
+    pub threshold: u8,
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, Default)]
+pub struct GenerateAccountResponse {
+    pub accounts:  Vec<PayloadAccount>,
+    pub threshold: u8,
```
huwenchao, line 37: `threshold` and `accounts` are useless in response.

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+    ) -> ServiceResponse<VerifyResponse> {
+        let wit_str = String::from_utf8(payload.witness.to_vec());
+        if wit_str.is_err() {
+            return ServiceResponse::<VerifyResponse>::from_error(
+                112,
+                "witness from utf8 bytes to string failed".to_owned(),
+            );
+        }
+
+        let wit_res: Result<Witness, _> = serde_json::from_str(wit_str.unwrap().as_str());
```
huwenchao, line 42: 1. There is [`serde_json::from_slice`](https://docs.serde.rs/serde_json/fn.from_slice.html) for your use case;
2. The witness in payload is Bytes, it means you may have to pass the payload as `{"witness":[1,2,3]}` to call this function. You may want to use `String` as the witness type here.

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+        if wit_res.is_err() {
+            return ServiceResponse::<VerifyResponse>::from_error(
+                112,
```
huwenchao, line 45: Error code is duplicated. Is it on purpose? I recommend to use different error codes.

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                accounts:  Vec::<Account>::new(),
+                threshold: 0,
+            });
+
+        if permission.threshold == 0 {
+            return ServiceResponse::<VerifyResponse>::from_error(
+                110,
+                "account not existed".to_owned(),
+            );
+        }
```
huwenchao, line 82: Why use `unwrap_or` and check for `threshold`? IMO, check whether the result of `get_account_value` is `None` or `Some` is better.

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+
+        let mut weight_sum = 0;
+        let size = permission.accounts.len();
+        let mut hash_account = [false; MAX_PERMISSION_ACCOUNTS as usize];
+
+        for i in 0..wit.signatures.len() {
+            let res = verify_single_sig(&payload.tx_hash, &wit.signatures[i], &wit.pubkeys[i]);
```
huwenchao, line 89: If length of pubkeys is less than signatures, it may panic here.

built-in-services/account/src/types.rs
fendouhyz: Does not outside client or muta browser  need this?

built-in-services/account/src/lib.rs
fendouhyz: ok, I will add  length check

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                110,
+                "accounts weight or threshold not valid".to_owned(),
+            );
+        }
+
+        let tx_hash = ctx.get_tx_hash().unwrap();
+
+        let keccak = HasherKeccak::new();
+        let addr_hash = Hash::from_bytes(Bytes::from(keccak.digest(&tx_hash.as_bytes())));
+        if addr_hash.is_err() {
```
huwenchao, line 195: This is not a user error, and should never happen. You can just use `let addr_hash = Hash::from_bytes(Bytes::from(keccak.digest(&tx_hash.as_bytes()))).except("some meaningful msg");` to make it panic here.

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+            weight_all += item.weight;
+            accounts.push(Account {
+                address:       item.address.clone(),
+                account_type:  ACCOUNT_TYPE_PUBLIC_KEY,
+                permission_id: 0,
+                weight:        item.weight,
+            });
+        }
+
+        if weight_all < payload.threshold || payload.threshold == 0 {
```
huwenchao, line 184: `payload.threshold == 0` will never happen after `weight_all < payload.threshold`.

built-in-services/account/src/lib.rs
fendouhyz: actually , check for threshold  is the alternative coding for None check.
I think your advise is right, i will rewrite it your way

built-in-services/account/src/lib.rs
fendouhyz: fixed

built-in-services/account/src/lib.rs
fendouhyz: I am not sure in some case, e.g. tests , inner call or other way direct call ,maybe tx_hash is not valid

built-in-services/account/src/types.rs
huwenchao: We use `event` in this case. If we invoke `generate_account` method from other service or contract, we can only know the change from event. If we do need them, we can get them from transaction payload.

built-in-services/account/src/lib.rs
huwenchao: Panic inside service will not crash the chain. The framework will catch the panic and stop the executing process instead of generating a transaction failure and writing it into the receipt. So it is better to panic for unexpected system error.

fendouhyz: ok

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                110,
+                "accounts weight or threshold not valid".to_owned(),
+            );
+        }
+
+        let tx_hash = ctx.get_tx_hash().unwrap();
+
+        let keccak = HasherKeccak::new();
+        let addr_hash = Hash::from_bytes(Bytes::from(keccak.digest(&tx_hash.as_bytes())));
```
yejiayu, line 194: let muil_addr = Address ::from_hash(Hash::digest(tx_hash.as_bytes()))

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+        };
+
+        ServiceResponse::<GenerateAccountResponse>::from_succeed(response)
+    }
+}
+
+fn verify_single_sig(tx_hash: &Hash, sig: &Hex, pubkey: &Hex) -> ServiceResponse<VerifyResponse> {
+    let data_hash = tx_hash.as_bytes();
+
+    let data_sig = hex::decode(sig.as_string_trim0x());
```
yejiayu, line 231: The conversion is not necessary, try to use `tx_hash.as_bytes().to_vec()`

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+            );
+        }
+
+        let mut weight_sum = 0;
+        let size = permission.accounts.len();
+        let mut hash_account = [false; MAX_PERMISSION_ACCOUNTS as usize];
+
+        for i in 0..wit.signatures.len() {
+            let res = verify_single_sig(&payload.tx_hash, &wit.signatures[i], &wit.pubkeys[i]);
+            if res.is_error() {
```
yejiayu, line 90: Why continue if the validation fails?

built-in-services/account/src/types.rs
fendouhyz: ok

built-in-services/account/src/lib.rs
fendouhyz: multiSig passed  means the sum of verified signatures reaches the threshold, whether other signatures are verified or not

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                return ServiceResponse::<VerifyResponse>::from_error(
+                    113,
+                    "len of signatures or pubkeys must be 1".to_owned(),
+                );
+            }
+            return verify_single_sig(&payload.tx_hash, &wit.signatures[0], &wit.pubkeys[0]);
+        }
+
+        // multi sig
+        if wit.signature_type != ACCOUNT_TYPE_MULTI_SIG {
```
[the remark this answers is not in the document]
fendouhyz: it will reture error response with "signature_type not valid"

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+        let wit_res: Result<Witness, _> = serde_json::from_str(wit_str.unwrap().as_str());
+        if wit_res.is_err() {
+            return ServiceResponse::<VerifyResponse>::from_error(
+                112,
+                "witness not valid".to_owned(),
+            );
+        }
+
+        let wit = wit_res.unwrap();
+        if wit.signature_type == ACCOUNT_TYPE_PUBLIC_KEY {
```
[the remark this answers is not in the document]
fendouhyz: if you don't use multiSig ever, you'll be confused about SINGLE_SIGN.
ACCOUNT_TYPE_PUBLIC_KEY means the account is generated from public_key, it may be natural for blockchain user?

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    #[cycles(100_00)]
+    #[read]
+    fn verify_signature(
+        &self,
+        ctx: ServiceContext,
+        payload: VerifyPayload,
+    ) -> ServiceResponse<VerifyResponse> {
+        let wit_str = String::from_utf8(payload.witness.to_vec());
+        if wit_str.is_err() {
```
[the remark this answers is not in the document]
fendouhyz: `Response::forbidden()` , use enum or const?

built-in-services/account/src/lib.rs
fendouhyz: ok

built-in-services/account/src/lib.rs
fendouhyz: ok

built-in-services/account/src/lib.rs
fendouhyz: ok

homura:
There are so many error codes, it seems that the framework needs to provide a centralized error code management module to avoid error code conflicts. Also, this may be helpful for error code location and automated documentation

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+            return ServiceResponse::<VerifyResponse>::from_error(
+                117,
+                "account not existed".to_owned(),
+            );
+        }
+
+        let permission = permission_res.unwrap();
+        let mut weight_sum = 0;
+        let size = permission.accounts.len();
+        let mut hash_account = [false; MAX_PERMISSION_ACCOUNTS as usize];
```
mkxbl, line 86: Dose `hash_account` mean has_account ?

built-in-services/account/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+            });
+
+        if permission.threshold == 0 {
+            return ServiceResponse::<GenerateAccountResponse>::from_error(
+                110,
+                "account not existed".to_owned(),
+            );
+        }
+
+        let mut accounts = Vec::<PayloadAccount>::new();
```
mkxbl, line 140: It seems that variable `accounts` is not related to the method response ?

built-in-services/account/src/lib.rs
fendouhyz: yes, has_account seems like a better name

built-in-services/account/src/lib.rs
fendouhyz: right, response is modified,i  forgot to delete the accounts

PR #259, opened 2020-04-28
fix(sync): Avoid requesting redundant transactions

yejiayu:
**What type of PR is this?**
fix

Fixed redundant requests while synchronizing blocks.

1. https://github.com/nervosnetwork/muta/pull/259#discussion_r416545078
2. https://github.com/nervosnetwork/muta/pull/259#discussion_r416545662

Fixes #

### DIFF
--- a/core/consensus/src/synchronization.rs
+++ b/core/consensus/src/synchronization.rs
@@ -150,8 +150,6 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
     ) -> ProtocolResult<()> {
         let mut current_consented_height = current_height;
 
-        let mut prepared_rich_block: Option<RichBlock> = None;
-
         while current_consented_height < remote_height {
             let consenting_height = current_consented_height + 1;
             log::info!(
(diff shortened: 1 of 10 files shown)


core/consensus/src/synchronization.rs
```diff
@@ -160,47 +158,28 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
...
-                        );
-                        e
-                    })?,
-
-                Some(_) => prepared_rich_block.take().unwrap(),
-            };
-
-            let consenting_proof: Proof = if consenting_height < remote_height {
-                let proof_block = self
-                    .get_rich_block_from_remote(ctx.clone(), consenting_height + 1)
```
yejiayu, line 180: Need proof instead of the full block.

core/consensus/src/synchronization.rs
```diff
@@ -263,11 +242,7 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
                 .verify_txs_sync(
                     ctx.clone(),
                     consenting_height,
-                    consenting_rich_block
```
yejiayu, line 266: Enough transactions have been requested so that there is no need to request them again.

PR #264, opened 2020-04-29
feat(framework): auto-gen service schemas + upgrade event mechanism

mkxbl:
**What type of PR is this?**

feat(framework)

1. Auto generate schema(graphql) of service

- use `#[derive(SchemaObject)]` to mark payload or response of service method
- use `#[description("comments")]` to add comments to schema
```rust
   #[derive(SchemaObject)]
   #[description("This is TestA")]
   struct TestA {
        #[description("This is String a")]
        a: String,
        #[description("This is TestB b")]
        b: TestB,
        #[description("This is bool c")]
        c: bool,
    }
    #[derive(SchemaObject)]
    #[description("This is TestB")]
    struct TestB {
        #[description("This is Vec<u8> d")]
        d: Vec<u8>,
    }
```
- use `#[derive(SchemaEvent)]` to mark all events struct
```rust
#[derive(SchemaEvent)]
pub enum Event {
    Asset,
    TransferEvent,
    ApproveEvent,
    TransferFromEvent,
}
```

- use `#[service(Event)]` to inform framework that your service will emit events, which is optional

2. Add topic attribute to `Event` and remove bloom

3. Add macro `emit_event!` to emit event

4. Fix: clear events if tx exec failed

5. Add new chain-api to get schema 

```graphql
type Query {
  getSchema: ChainSchema!
}

type ChainSchema {
  schema: [ServiceSchema!]!
}

type ServiceSchema {
  service: String!
  method: String!
  event: String!
}
```

Fixes #


framework/src/executor/mod.rs
```diff
@@ -59,6 +58,31 @@ impl<S: Storage, DB: TrieDB, Mapping: ServiceMapping> Clone for ServiceExecutor<
...
+            );
+            let service = mapping.get_service(&name, sdk)?;
+            let ret = service.schema_();
+            schema.push(ServiceSchema {
+                service: name.clone(),
+                method:  ret.0,
+                event:   ret.1,
+            });
+        }
+        storage.insert_schema(ChainSchema { schema }).await
```
huwenchao, line 83: Seems like it is not necessary to insert the schema into storage. Is is possible to remain it in memory, like a global lazy static variable, and make it accessible for API module? Then we don't need to implement the `insert_schema` and `get_schema` method.

binding-macro/src/schema.rs
```diff
@@ -0,0 +1,227 @@
...
+                } else {
+                    panic!("#[derive(SchemaObject)]: field type only supports T, Vec<T>")
+                }
+            } else {
+                panic!("#[derive(SchemaObject)]: length of field type should be 1")
+            }
+        }
+        _ => panic!("#[derive(SchemaObject)]: field type only supports T, Vec<T>"),
+    }
+}
```
huwenchao, line 216: It is hard to read for the deep indent. Try to make it flatter.

yejiayu:
Schema considers the layered design.

schema_derive only returns a struct describing metadata and is not responsible for direct conversion to GQL language.

The conversion to GQL language will be done in Muta-API.

framework/src/executor/mod.rs
mkxbl: That is really a option, and i think it twice. But there's no obvious benefits to change to that, maintaining it in memory, because we don't use schema frequently while running node.

binding-macro/src/schema.rs
mkxbl: In this circumstance, i think it's hard to do that. Can you provide some specifics ?

huwenchao: 1. Move the `else` part forward.

```
// change
if xx {
  // long code
} else {
  panic!
}

// to
if !xx {
  panic
}
// long code
```

2. Extract the inner variable first and then write the long logic code, other than writing it inside the block. 

```
// change
match ty {
    Type::Path(ty) => {
       // long code
    },
    _ => panic!("#[derive(SchemaObject)]: field type only supports T, Vec<T>"),
};

// to
let inner_ty = match ty {
    Type::Path(ty) => ty,
    _ => panic!("#[derive(SchemaObject)]: field type only supports T, Vec<T>"),
};
// long code using inner_ty
```

yejiayu: @huwenchao  You still owe me a task, the muta-bot's weekly report isn't finished yet  🤪

framework/src/executor/mod.rs
yejiayu: I think keeping it in memory is a better option, with the benefit that no matter how long the program runs, only the first call is expensive.

huwenchao: The usage of schema in storage is a little weird. We insert it into storage every time we start the process, and get it when the API invoke get_schema. For me it's more like a memory thing than a storage thing. The drawback is that you have to implement `insert_schema` and `get_schema` method in storage trait.

Consider that you have already implemented it in this way, it also makes sense. The performance differs little. So it is not a big deal and pretty fine to remain it in this way.

yejiayu: When you consider [this design](https://github.com/nervosnetwork/muta/pull/264#issuecomment-621670185), you will feel that it is a natural thing to keep in memory.

binding-macro/src/schema.rs
mkxbl: Nice move, I will change to that

framework/src/executor/mod.rs
mkxbl: Ok, i will keep it in memory

mkxbl:
> Schema considers the layered design.
> 
> `schema_derive` only returns a struct describing metadata and is not responsible for direct conversion to GQL language.
> 
> The conversion to GQL language will be done in Muta-API.

Done

yejiayu:
/close

until framewrok2.0

PR #271, opened 2020-05-07
feat(metric): introduce metric using prometheus

zeroqn:
**What type of PR is this?**
feat

Introduce metric lib

Fixes #


core/api/src/lib.rs
```diff
@@ -142,13 +143,25 @@ impl Mutation {
...
+            .await
+        {
+            if err.to_string().contains("already commit") {
+                common_metrics::api::REPEATED_TX_COUNT.inc();
+            }
+            return Err(err.into());
+        }
+
+        common_metrics::api::SUCCESS_TX_COUNT.inc();
+        common_metrics::api::TX_SUCCESS_TIME_COST.observe_duration(inst.elapsed());
```
homura, line 164: the [timer](https://github.com/tikv/rust-prometheus/blob/master/examples/example_hyper.rs#L40-L53) seems more convenient

zeroqn: Yes, but we changed to use static histogram now, which seems don't have this func.

PR #292, opened 2020-05-15
refactor(mempool): Avoid duplicate validation transactions when sync

yejiayu:
**What type of PR is this?**
refactor

If the transaction is already in the `mempool` cache, we can skip validation.

This will greatly improve the synchronization efficiency when height - 1.

Fixes #

### DIFF
--- a/core/mempool/src/lib.rs
+++ b/core/mempool/src/lib.rs
@@ -255,23 +255,27 @@ where
         tx_hashes: Vec<Hash>,
     ) -> ProtocolResult<Vec<SignedTransaction>> {
         let len = tx_hashes.len();
+        let mut missing_hashes = vec![];
         let mut full_txs = Vec::with_capacity(len);
 
-        for tx_hash in &tx_hashes {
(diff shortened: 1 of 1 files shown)


core/mempool/src/lib.rs
```diff
@@ -255,23 +255,27 @@ where
         tx_hashes: Vec<Hash>,
     ) -> ProtocolResult<Vec<SignedTransaction>> {
         let len = tx_hashes.len();
+        let mut missing_hashes = vec![];
         let mut full_txs = Vec::with_capacity(len);
 
-        for tx_hash in &tx_hashes {
+        for tx_hash in tx_hashes.iter() {
```
rev-chaos, line 261: It's impossible that the query transactions in both memory and storage. So this modification can be confusing.

yejiayu: This requires you to get the lock first (flush_lock), but `pull_txs` is an externally accessible interface and is accessed very frequently, and getting a mutex each time can significantly block the process.

PR #293, opened 2020-05-26
refactor(store): change store map to iter

KaoImin:
**What type of PR is this?**
refactor

change the vector into buckets to iter

Fixes #


framework/src/binding/store/map_new.rs
```diff
@@ -0,0 +1,393 @@
...
+use rayon::prelude::*;
+
+use protocol::fixed_codec::FixedCodec;
+use protocol::traits::{ServiceState, StoreMap};
+use protocol::types::Hash;
+use protocol::{ProtocolError, ProtocolResult};
+
+use crate::binding::store::{get_bucket_index, Bucket, FixedBuckets, StoreError};
+
+pub struct NewStoreMap<S: ServiceState, K: FixedCodec + PartialEq, V: FixedCodec> {
```
mkxbl: Maybe the name "NewXXX" of struct  not customary ?

PR #294, opened 2020-05-27
feat: supported new metrics

homura:
**What type of PR is this?**
feat

add some metrics for Prometheus

Fixes #


core/consensus/src/engine.rs
```diff
@@ -31,6 +31,7 @@ use crate::message::{
     END_GOSSIP_AGGREGATED_VOTE, END_GOSSIP_SIGNED_CHOKE, END_GOSSIP_SIGNED_PROPOSAL,
     END_GOSSIP_SIGNED_VOTE,
 };
+use crate::metrics::*;
```
zeroqn, line 34: Don't use * import.

homura: improved

PR #295, opened 2020-05-28
refactor(sync)!: Use signed transaction hashes to skip transaction signature verification

yejiayu:
**What type of PR is this?**
refactor

After the current synchronization to the complete block, it is still necessary to check the signature of the transaction, the legality of the transaction, and the double-check of the database.

These tasks are actually unnecessary. As long as the proof of the synchronized block (note: this proof does not refer to the proof contained in the block header, but the proof of the block) passes the check, then it is equivalent to the hash of the block protected by the proof is Trusted, so the content of the block protected by the block header is trusted. That being the case, why bother checking the transaction again. You only need to verify the integrity of the content (that is, you need to check whether the hash of the transaction matches and whether the Merkle root of the calculation transaction matches the header).

One point to note here is that since the transaction hash does not contain a signature, the transaction signature of the synchronized block is not protected by the block header, and its integrity and correctness cannot be guaranteed. One of the solutions is to add a field in the block header-the the complete transaction hash, which is the hash value calculated for all complete transactions as a whole.

Fixes #

See comments


protocol/src/types/block.rs
```diff
@@ -33,22 +34,22 @@ pub struct Block {
...
-    pub proposer:          Address,
-    pub proof:             Proof,
+    pub chain_id: Hash,
+    pub height: u64,
+    pub exec_height: u64,
+    pub prev_hash: Hash,
+    pub timestamp: u64,
+    pub logs_bloom: Vec<Bloom>,
+    pub order_root: MerkleRoot,
+    pub order_signed_transactions_hash: Hash,
```
yejiayu, line 44: Add the order_signed_transactions_hash to BlockHeader

core/consensus/src/engine.rs
```diff
@@ -78,13 +78,18 @@ impl<Adapter: ConsensusAdapter + 'static> Engine<FixedPill> for ConsensusEngine<
...
                 current_consensus_status.cycles_limit,
                 current_consensus_status.tx_num_limit,
             )
             .await?
             .clap();
+        let signed_txs = self
+            .adapter
+            .get_full_txs(ctx.clone(), ordered_tx_hashes.clone())
+            .await?;
+        let order_signed_transactions_hash = digest_signed_transactions(&signed_txs)?;
```
yejiayu, line 92: The hash value of the signed transaction is calculated at the time of the proposal.

core/consensus/src/synchronization.rs
```diff
@@ -249,17 +250,24 @@ impl<Adapter: SynchronizationAdapter> OverlordSynchronization<Adapter> {
...
-                    ctx.clone(),
-                    consenting_height,
-                    consenting_rich_block.txs.clone(),
-                )
-                .await
-                .map_err(|e| {
-                    log::error!("[synchronization]: verify_txs_sync error",);
-                    e
-                })?;
+            let order_signed_transactions_hash =
```
yejiayu, line 253: Skip transaction signature verification

PR #315, opened 2020-06-11
feat(api)!: Allowed to return None

yejiayu:
**What type of PR is this?**
feat

Returns `None` instead of the `not found` error if the resource does not exist.

Affected
- get_receipt_by_tx_hash
- get_transaction_by_hash
- get_block

Fixes #


yejiayu:
@homura 
This is a breaking change that requires SDK adaptation.

yejiayu:
@homura 
Has The JS-SDK a new version of adaptation?

PR #318, opened 2020-06-11
feat(service)!: add multiple signatures service

KaoImin:
**What type of PR is this?**
feature

Develop multiple signatures service.
The multiple signatures account is consists of several addresses, a threshold, a memo string, and an owner. Each address has a corresponding weight. As long as the sum of the signed weights exceeds the threshold, the signature will be effective. The owner can change the multiple signature account includes adding an account, removing an account, changing the weight, changing the threshold, and so on. The owner can also be changed after being generated. Last but not least, the nesting of multiple signatures in multiple signatures is allowed.

Notice that: 
* The maximum count of signatures in multiple signatures is 16.
* The maximum recursion depth of multiple signatures is 8.

Break change:
* The `pubkey` and `signature` field in `SignedTransaction` struct has been changed. Both of these are serialized from a`Vec<T>`.

Fixes #


built-in-services/multi-signature/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        )) {
+            let accounts = payload
+                .addr_with_weight
+                .iter()
+                .map(|item| Account {
+                    address:     item.address.clone(),
+                    weight:      item.weight,
+                    is_multiple: !self
+                        .get_account_from_address(ctx.clone(), GetMultiSigAccountPayload {
+                            multi_sig_address: item.address.clone(),
```
LycrusHamster, line 75: check recursive

built-in-services/multi-signature/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+            &mut recursion_depth,
+        )
+    }
+
+    fn _verify_multi_signature(
+        &self,
+        tx_hash: &Hash,
+        wit_map: &HashMap<Address, (Bytes, Bytes)>,
+        sender: &Address,
+        weight_acc: &mut u32,
```
LycrusHamster, line 173: the weight_acc is useless

yejiayu:
if a breaking change, you should use `!` symbol or `BREAKING CHANGE` to the PR title.
eg.
`feat(service)!: add ... or BREAKING CHANGE: add ...`

https://www.conventionalcommits.org/en/v1.0.0

built-in-services/multi-signature/src/lib.rs
```diff
@@ -0,0 +1,554 @@
...
+                        .get_account_from_address(ctx.clone(), GetMultiSigAccountPayload {
+                            multi_sig_address: item.address.clone(),
+                        })
+                        .is_error(),
+                })
+                .collect::<Vec<_>>();
+
+            let permission = MultiSigPermission {
+                accounts,
+                owner: payload.owner,
```
cosinlink, line 137: owner  also has the risk of MAX_MULTI_SIGNATURE_RECURSION_DEPTH  overflow

built-in-services/multi-signature/src/lib.rs
```diff
@@ -0,0 +1,554 @@
...
+            if let Ok(addr) = Address::from_pubkey_bytes(pubkeys[0].clone()) {
+                if addr == payload.sender {
+                    return self._verify_single_signature(
+                        &ctx.get_tx_hash().unwrap(),
+                        &signatures[0],
+                        &pubkeys[0],
+                    );
+                }
+            } else {
+                return ServiceResponse::<()>::from_error(123, "invalid public key".to_owned());
```
cosinlink, line 207: it should not return error when addr != payload.sender in the case of 'pubkeys.len() == 1'. Because it may be a multiSig address while pubkeys.len() == 1,  pubkeys.len is just lenght of signature.

built-in-services/multi-signature/src/lib.rs
```diff
@@ -0,0 +1,554 @@
...
+            &mut recursion_depth,
+        )
+    }
+
+    fn _verify_multi_signature(
+        &self,
+        tx_hash: &Hash,
+        wit_map: &HashMap<Address, (Bytes, Bytes)>,
+        sender: &Address,
+        recursion_depth: &mut u8,
```
cosinlink, line 225: It should not be &mut u8 here.
In this case, recursion_depth recorded the num of all multiSig accounts in the deep search tree, not the depth.

PR #320, opened 2020-06-22
feat(service)!: add authorization service for mempool

KaoImin:
**What type of PR is this?**
feature

Add authorization service.
After adding the multi-signature service, the mempool will not be able to verify the signature of the transaction. Besides, there are some checking items that should be checked before inserting a transaction such as check the block list, check the account balance. This PR adds a service named authorization. Every check item except verification signature should be registered in genesis. The content in genesis is the name string of the service and the name string of the method. To be noticed that the payload type of the method must be `protocol::types::SignedTransaction`.

Fixes #


LycrusHamster:
register multisig in DefaultServiceMapping

built-in-services/authorization/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        for item in service_names.into_iter().zip(function_names.into_iter()) {
+            self.verified_map.insert(item.0, item.1);
+        }
+    }
+
+    #[cycles(210_00)]
+    #[read]
+    fn check_authorization(&self, ctx: ServiceContext, payload: String) -> ServiceResponse<()> {
+        for (service_name, mathod_name) in self.verified_map.iter() {
```
LycrusHamster, line 48: spelling error mathod -> method

PR #322, opened 2020-06-29
refactor(service): calculate multi-signature recursion depth

KaoImin:
**What type of PR is this?**
refactor

* Instead of using a global variable, use a local variable to calculate the multi-signature recursion depth.
* remove redundancy signature verify in `add_account`, `remove_account`, `set_threshold` .etc
* refactor check recursion depth function, when the depth is overflow, stop check immediately.
* Add adaptive address. If the owner address in the payload is `0xffffffffffffffffffffffffffffffffffffffff`, use the multi-signature address as the owner address.

Fixes #

### DIFF
--- a/built-in-services/multi-signature/Cargo.toml
+++ b/built-in-services/multi-signature/Cargo.toml
@@ -15,6 +15,7 @@ common-crypto = { path = "../../common/crypto" }
 derive_more = "0.99"
 hasher = { version="0.1", features = ["hash-keccak"] }
 hex = "0.4"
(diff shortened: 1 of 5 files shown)


LycrusHamster:
修复ci

https://github.com/nervosnetwork/Huobi-doc-specs 检查一下有没有没有对的上的

    fn update_account(new_config: MultiSigPermission) -> bool {} //更新多签地址的信息
增加一个这个接口,你那些接口,用着太难受了.....

PR #337, opened 2020-07-06
refactor(auth): change verified map to array

KaoImin:
**What type of PR is this?**
refactor

* Change the type of `verified_map` from `StoreMap` to `StoreArray`, so that different methods of the same service can be registered.
* Remove `verify_signature` method is forbidden.

Fixes #


LycrusHamster:
I have no idea.
I have a peek and I think it is just a refactor for better impl

KaoImin:
> /cc @LycrusHamster @KaoImin
> 
> What does this PR change have to do with?

It just makes the authorization service better.

PR #347, opened 2020-07-08
feat!: add 'name' field in event struct

LycrusHamster:
**What type of PR is this?**
BREAKING CHANGE

feature

to enhance event to include one new field:
name : String , this indicates the event name in the context

provided applied, this pr will help client to use 'name' field to filter events they interest, instead of try read JSON data to identify JSON data's type

### DIFF
--- a/built-in-services/asset/src/lib.rs
+++ b/built-in-services/asset/src/lib.rs
@@ -173,7 +173,7 @@ impl<SDK: ServiceSDK> AssetService<SDK> {
             return ServiceResponse::<Asset>::from_error(103, format!("{:?}", e));
         }
         let event_str = event_res.unwrap();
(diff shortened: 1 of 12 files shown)


yejiayu:
plz follow https://www.conventionalcommits.org/en/

built-in-services/asset/src/lib.rs
```diff
@@ -173,7 +173,7 @@ impl<SDK: ServiceSDK> AssetService<SDK> {
             return ServiceResponse::<Asset>::from_error(103, format!("{:?}", e));
         }
         let event_str = event_res.unwrap();
-        ctx.emit_event(event_str);
+        ctx.emit_event("Create".to_owned(), event_str);
```
homura, line 176: Consistency with the method name(`create_asset`) will make it clearer.

PR #348, opened 2020-07-09
refactor!(network): rename users_cast to multicast, pass peer_ids instead

zeroqn:
**What type of PR is this?**
refactor

refactor network api
  - rename users_cast t omulticast
  - replace Vec<Address> with Vec<Bytes>, PeerId bytes

Fixes #


zeroqn:
Split this PR. #354 first.

PR #352, opened 2020-07-13
test(e2e): add more cases of multi-signature service

KaoImin:
**What type of PR is this?**
test

Fixes #


LycrusHamster:
@KaoImin  why change defaul 1337 p2p port to 13390?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
yejiayu, line 1: the `yarn.lock` should not be in root dir

devtools/chain/config.toml
```diff
@@ -13,7 +13,7 @@ maxconn = 25000
 max_payload_size = 1048576
 
 [network]
-listening_address = "0.0.0.0:1337"
+listening_address = "0.0.0.0:13390"
```
yejiayu, line 16: why

core/mempool/src/lib.rs
```diff
@@ -419,7 +419,7 @@ pub enum MemPoolError {
     InsertCandidate { len: usize },
 
     #[display(fmt = "Tx: {:?} check_sig failed", tx_hash)]
-    CheckSig { tx_hash: Hash },
+    CheckAuthorization { tx_hash: Hash },
```
yejiayu, line 422: if you change to `CheckAuthorization` that error message should be changed as well.

devtools/chain/config.toml
KaoImin: The "0.0.0.0:1337" is always occupied.

PR #355, opened 2020-07-14
feat(framework)!: Add return type ServiceResponse<String> for tx_hook

LycrusHamster:
**What type of PR is this?**

feat!: add tx_hook return type

this could imply the result of tx_hook

### DIFF
--- a/binding-macro/tests/mod.rs
+++ b/binding-macro/tests/mod.rs
@@ -22,14 +22,15 @@ use protocol::types::{
 fn test_read_and_write() {
     struct Tests;
 
+    #[service]
     impl Tests {
         #[read]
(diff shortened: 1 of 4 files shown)


binding-macro/tests/mod.rs
```diff
@@ -74,6 +74,31 @@ fn test_hooks() {
...
 
+#[test]
+fn test_tx_hooks() {
+    struct Tests {
+        pub height: u64,
+    };
+
+    impl Tests {
+        #[tx_hook_after]
+        fn tx_hook_after(&mut self, params: &ExecutorParams) {
```
yejiayu, line 85: Is this case out of date?

LycrusHamster: fixed

PR #365, opened 2020-07-20
perf(executor): use inner call instead of service dispatcher

KaoImin:
**What type of PR is this?**
perf

1. Add an `SDKFactory` trait to cache SDKs of each service at the beginning of executing a block.
2. Remove service dispatcher.
3. The check items of the authorization service are hard-coded into the code and cannot be changed through transactions.

```
Bench in Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz (8 x 2200)
100 txs bench_execute ... bench:  11,299,912 ns/iter (+/- 3,402,276)
1000 txs bench::bench_execute ... bench: 101,187,934 ns/iter (+/- 26,000,469)
```

Fixes #


yejiayu:
Add describe

yejiayu:
I think you should define a `trait` for functions if you don't want to be exposed, and give you two examples

   1. metadata

   ```rust
   pub trait MetadataService {
     fn get(&self) -> Metadata;
     fn update(&mut self, metadata: Metadata);
   }
   
   #[service]
   impl NormalMetadataService {
     #[read]
     // the function can be called with transaction
     fn get_metadata() -> Metadata { self.get() }
   }
   
   impl NormalMetadataService for MetadataService {
     // the function can be called only each other services
     fn get() -> Metadata { ... }
     
     // Now, we has a safety calling way on updating metadata that don't need rewrite metadata.
     fn update(&mut self) -> { ... }
   }
   ```

2. Authorization

   ```rust
   pub trait AuthorizationService {
     fn allow(&self) -> bool;
   }
   
   // We don't need to expose any methods.
   impl NoopAuthorizationService for AuthorizationService {
     fn allow(&self) -> { ... };
   }
   ```

Cargo.toml
```diff
@@ -81,5 +85,5 @@ crate-type = ["bin"]
 
 [[test]]
 name = "trust_metric"
-path = "tests/trust_metric.rs"
+path = "tests/tests.rs"
```
yejiayu, line 88: Why did you change the path here?

KaoImin: The name `trust_metric` is not suitable, and I rename it.

yejiayu: @zeroqn

zeroqn: IMO, ```test.rs``` means a general test, but ```[[test]]``` here means a specific
integration test.

Each integration test results in a separate executable binary. 
reference: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#tests

benchmark/benchmark_genesis.toml
```diff
@@ -0,0 +1,52 @@
...
+                "discount_percent": 50
+            }
+        ],
+        "miner_benefit": 10
+    },
+    "tx_fee_inlet_address": "0x503492f4bddc731a72b8caa806183f921c284f8e",
+    "miner_profit_outlet_address": "0xcff1002107105460941f797828f468667aa1a2db",
+    "miner_charge_map": []
+}
+'''
```
yejiayu, line 52: new line

built-in-services/asset/src/lib.rs
```diff
@@ -322,6 +331,37 @@ impl<SDK: ServiceSDK> AssetService<SDK> {
         ServiceResponse::<()>::from_succeed(())
     }
 
+    #[cycles(210_00)]
+    #[write]
+    pub fn hook_transfer_from(
```
yejiayu, line 336: If the service asset is spending cycle fee, our benchmark script maybe failure.

PR #371, opened 2020-07-23
fix(consensus): return an error when committing an outdated block

KaoImin:
**What type of PR is this?**
fix

Return an error when committing an outdated block, avoid panic here.

Fixes #


yejiayu:
/hold 

unit test

PR #377, opened 2020-07-27
fix(consensus): Add timestamp checking

yejiayu:
**What type of PR is this?**
fixed a potential security problem.

If the timestamp of the proposal is not verified, the bad node will be able to pass an arbitrarily large time, causing a huge fluctuation in the blockchain time.

So, we should limit the proposal's timestamp to two conditions:
1. The previous block timestamp should be less than proposal timestamp.
2. this node timestamp should be longer than the proposal timestamp

Fixes #

### DIFF
--- a/core/consensus/src/adapter.rs
+++ b/core/consensus/src/adapter.rs
@@ -1057,12 +1057,12 @@ mod tests {
         assert_eq!(validate_timestamp(10, 11, consensus_interval), false);
 
         // current 10 == proposal 10. true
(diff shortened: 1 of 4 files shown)


yejiayu:
/hold

There has been some degradation in performance.

image
image

yejiayu:
/hold cancel

The performance returned to normal.
image

/cc @rev-chaos @KaoImin

yejiayu:
ping @KaoImin @rev-chaos

yejiayu:
ping @KaoImin @rev-chaos

PR #380, opened 2020-07-29
feat(network): split transmitter data

zeroqn:
**What type of PR is this?**
feat

This PR introduce a feature that automatically split data into chunks base on hardcode value in transmitter protocol. And we are also ensure data sending in order, so chunks from different data are not interweaved with
each other.

Fixes #


core/network/src/protocols/transmitter/protocol.rs
```diff
@@ -0,0 +1,100 @@
...
+            Some(pk) => pk.peer_id(),
+            None => {
+                // Dont care result here, connection/keeper will also handle this.
+                let _ = ctx.disconnect(ctx.session.id);
+                return;
+            }
+        };
+        let session_id = ctx.session.id;
+
+        if data.len() < 10 {
```
yejiayu, line 41: 10 = Minimum protocol package?

yejiayu:
Reviewed the core logic of splitting data, LGTM

core/network/src/protocols/transmitter/protocol.rs
zeroqn: ```seq``` takes 8 bytes, ```bool``` takes 1 byte, so a valid data length must be bigger or equal than 10.

PR #397, opened 2020-08-04
perf(metrics): Add metrics of state

yejiayu:


common/apm/src/metrics/storage.rs
```diff
@@ -72,6 +73,20 @@ lazy_static! {
         auto_flush_from!(STORAGE_GET_CF_COUNTER_VEC, StorageGetCfTotalVec);
 }
 
+pub fn on_storage_get_state(duration: Duration, keys: i64) {
+    let seconds = duration_to_sec(duration);
+
+    STORAGE_PUT_CF_TIME_USAGE.state.inc_by(seconds);
```
homura, line 79: `get_state` is not `put_cf`, should be STORAGE_GET_CF_TIME_USAGE

PR #400, opened 2020-08-06
fix(apm): Return the correct time

yejiayu:
**What type of PR is this?**
fix

Return the correct time

Fixes #

### DIFF
--- a/common/apm/src/metrics.rs
+++ b/common/apm/src/metrics.rs
@@ -64,6 +64,7 @@ mod tests {
     fn test_duration_to_sec() {
         let d = Duration::from_millis(1110);
         let sec = duration_to_sec(d);
+
         assert_eq!(sec, 1.11 as f64);
     }
 }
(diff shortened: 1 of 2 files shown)


common/apm/src/metrics.rs
```diff
@@ -54,3 +54,18 @@ pub fn all_metrics() -> ProtocolResult<Vec<u8>> {
...
+    use std::time::{Duration, Instant};
+    use super::duration_to_sec;
+
+    #[test]
+    fn test_duration_to_sec() {
+        let inst = Instant::now();
+        let d = Duration::from_nanos(10000000010);
+        let sec = duration_to_sec(d);
+        let sec2 = duration_to_sec(inst.elapsed());
+        println!("sec {:?} sec2 {:?}", sec, sec2);
```
homura, line 69: `assert_eq` would be better, try to more cases like _2s_, _1.1s_, _0.11s_, _0.000000003s_

PR #404, opened 2020-08-07
perf(state): add state cache for trieDB

KaoImin:
Implement an approximately fixed size cache through HashMap. It caches the data used in the latest block, and it will be flushed when committing a block. The size of the cache is 2000, however, this does not mean that it can only store 2000 data. When the data in the cache is more than 2000 when the cache is refreshed, it will be randomly deleted to 2000. The random seed is the largest prime number within 50000, 49999.
 
Fixes #

Ref https://github.com/nervosnetwork/muta-docs/pull/55

No Breaking Change

### DIFF
--- a/framework/src/binding/state/trie_db.rs
+++ b/framework/src/binding/state/trie_db.rs
@@ -208,19 +208,31 @@ fn to_store_err(e: rocksdb::Error) -> RocksTrieDBError {
 }
 
 #[cfg(test)]
-mod benchmark {
+mod tests {
     extern crate test;
     use test::Bencher;
 
     use super::*;
 
     #[bench]
-    fn bench_insert(b: &mut Bencher) {
+    fn bench_rand(b: &mut Bencher) {
         b.iter(|| {
             let mut rng = SmallRng::seed_from_u64(49999);
             for _ in 0..10000 {
                 rng.gen_range(10, 1000000);
             }
         })
+    }
+
+    #[test]
+    fn test_rand_remove() {
+        let list = (0..10).collect::<Vec<_>>();
+        let keys = list.iter().collect::<Vec<_>>();
+        let to_removed_num = (1..10).collect::<Vec<_>>();
+
+        for num in to_removed_num.into_iter() {
+            let res = rand_remove_list(keys.clone(), num);
+            assert_eq!(res.len(), num);
+        }
     }
 }
--- a/framework/src/binding/state/trie_db.rs
+++ b/framework/src/binding/state/trie_db.rs
@@ -82,7 +82,12 @@ impl cita_trie::DB for RocksTrieDB {
         if res {
             Ok(true)
         } else {
-            Ok(self.db.get(key).map_err(to_store_err)?.is_some())
+            if let Some(val) = self.db.get(key).map_err(to_store_err)? {
+                let mut cache = self.cache.write();
+                cache.insert(key.to_owned(), val.to_owned());
+                return Ok(true);
+            }
+            Ok(false)
         }
     }
 
--- a/framework/src/binding/state/trie_db.rs
+++ b/framework/src/binding/state/trie_db.rs
@@ -1,17 +1,25 @@
+use std::collections::HashMap;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Instant;
 
 use bytes::Bytes;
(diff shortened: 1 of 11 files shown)


yejiayu:
Add the benchmark of the executor?  You need to make sure that it create a new MPT each time.

framework/src/binding/state/trie_db.rs
```diff
@@ -1,17 +1,25 @@
...
 use derive_more::{Display, From};
+use parking_lot::RwLock;
+use rand::{rngs::SmallRng, Rng, SeedableRng};
 use rocksdb::{Options, WriteBatch, DB};
 
 use common_apm::metrics::storage::{on_storage_get_state, on_storage_put_state};
 use protocol::{ProtocolError, ProtocolErrorKind, ProtocolResult};
 
+const RAND_SEED: u64 = 49999;
+const MAP_SIZE: usize = 2000;
```
yejiayu, line 16: we can extract it to the node config

framework/src/binding/state/trie_db.rs
```diff
@@ -36,20 +57,34 @@ impl cita_trie::DB for RocksTrieDB {
     fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
         let inst = Instant::now();
 
-        let res = self.db.get(key).map_err(to_store_err)?.map(|v| v.to_vec());
+        let res = self.inner_get(key)?;
 
         on_storage_get_state(inst.elapsed(), 1i64);
```
yejiayu, line 62: move it to really code of getting from the `RocksDB`.

framework/src/binding/state/trie_db.rs
```diff
@@ -26,8 +34,21 @@ impl RocksTrieDB {
...
     }
+
+    fn inner_get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, RocksTrieDBError> {
+        let res = {
+            let cache = self.cache.read();
+            cache.get(key).cloned()
+        };
+
+        if res.is_none() {
+            return Ok(self.db.get(key).map_err(to_store_err)?);
```
yejiayu, line 48: If not exists in that cache, should insert it after getting from the DB?  
Another question, what if is full of that cache at this moment?

framework/src/binding/state/trie_db.rs
KaoImin: The map size can be set in the config, however, the rand seed should hard code.

framework/src/binding/state/trie_db.rs
```diff
@@ -26,8 +34,21 @@ impl RocksTrieDB {
         Ok(RocksTrieDB {
             light,
             db: Arc::new(db),
+            cache: RwLock::new(HashMap::new()),
```
yejiayu, line 37: If the HashMap triggers resort, performance would suffer greatly because it would need to copy all the data once.

So, It is best to capacity for enough size in advance.

framework/src/binding/state/trie_db.rs
yejiayu: I just talking about `map_size`

framework/src/binding/state/trie_db.rs
KaoImin: If you insert the obtained data into the cache here, it is no different from LRU. Doing so cannot achieve true read concurrency.

framework/src/binding/state/trie_db.rs
KaoImin: ok

yejiayu:
Can you show a result of pref to this PR descriptions？

framework/src/binding/state/trie_db.rs
```diff
@@ -1,21 +1,33 @@
...
 use bytes::Bytes;
 use derive_more::{Display, From};
+use parking_lot::RwLock;
+use rand::{rngs::SmallRng, Rng, SeedableRng};
 use rocksdb::{Options, WriteBatch, DB};
 
 use common_apm::metrics::storage::{on_storage_get_state, on_storage_put_state};
 use protocol::{ProtocolError, ProtocolErrorKind, ProtocolResult};
 
+const RAND_SEED: u64 = 49999;
```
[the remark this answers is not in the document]
KaoImin: Nothing.

PR #405, opened 2020-08-09
fix: load hrp before deserializing genesis payload to take hrp effect

LycrusHamster:

fix: load hrp before deserializing genesis payload to take hrp effect

N/A

N/A

No Breaking Change

This is an internal bug and fix doesn't take any changes to toolchains and docs.

hrp should load firstly before deserializing genesis cause Address maybe in service genesis payload may need hrp

### DIFF
--- a/protocol/src/types/primitive.rs
+++ b/protocol/src/types/primitive.rs
@@ -14,6 +14,7 @@ use ophelia::UncompressedPublicKey;
 use ophelia_secp256k1::Secp256k1PublicKey;
 use serde::de;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 
 use crate::fixed_codec::{FixedCodec, FixedCodecError};
 use crate::types::TypesError;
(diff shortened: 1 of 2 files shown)


yejiayu:
We should add unit test for this PR which avoid running failures due to some breaking changes in future

LycrusHamster:
> We should add unit test for this PR which avoid running failures due to some breaking changes in future

So as I think. The problem is not to add test, but where to add?

yejiayu:
> > We should add unit test for this PR which avoid running failures due to some breaking changes in future
> 
> 
> 
> So as I think. The problem is not to add test, but where to add?

The point of testing is to match field, so i think it's just a matter of ensuring that the serialized Metadata json string can read the fields.

Can we add it to types/metadata.rs？

LycrusHamster:
> > > We should add unit test for this PR which avoid running failures due to some breaking changes in future
> > 
> > 
> > So as I think. The problem is not to add test, but where to add?
> 
> The point of testing is to match field, so i think it's just a matter of ensuring that the serialized Metadata json string can read the fields.
> 
> Can we add it to types/metadata.rs？

The hrp can only be set once, which means if I set hrp to sth. like 'meow', the following address's hrp have to be 'meow'
However cargo test is designed to run parelly and each test case must not relate to others.
I don't wan't to set test thread to 1.
And if I set hrp, other tests shall fail.

Thus.....you see the 'crippled' test cases...

PR #409, opened 2020-08-10
feat(test): Byzantine test

rev-chaos:
RT

Fixes #

Ref #

No Breaking Change


rev-chaos:
Have found 3 bugs so far:
1. multi-sig service rlp decode signature panic!  Kao fixed in https://github.com/nervosnetwork/muta/pull/411.
2. overlord rlp decode proposal panic! The cause is still being investigated.
3. overlord straightly inserts proposal which height equals current height but round is not current round without check, this will cause liveness problem in byzantine attack.

yejiayu:
Nice! add the byzantine test to ci?

byzantine/generators.toml
```diff
@@ -0,0 +1,448 @@
+interval = 500
```
yejiayu, line 1: Suggest splitting config for diff cases

byzantine/generators.toml
```diff
@@ -0,0 +1,448 @@
...
+probability = 1.0
+num_range = [1, 10]
+priority = "High"
+
+[[list]]
+req_end = "/rpc_call/mempool/pull_txs"
+msg_type = { PullTxs = "InvalidSender" }        # not cause bugs
+probability = 1.0
+num_range = [1, 10]
+priority = "High"
```
yejiayu, line 448: new line

KaoImin (approved):

PR #429, opened 2020-08-24
feat(apm): add executing block num to apm

KaoImin:
Add executing block number gauge to apm.

Fixes #

Ref #

No Breaking Change

### DIFF
--- a/common/apm/README.md
+++ b/common/apm/README.md
@@ -9,7 +9,7 @@ All current metrics and usage
 | muta_api_request_time_cost_seconds | histogram    |                          |
 
 
-## Consensus		
+## Consensus
 <table>
 <thead>
   <tr>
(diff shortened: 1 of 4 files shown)


yejiayu:
Don't forget the doc https://github.com/nervosnetwork/muta/blob/master/common/apm/README.md @KaoImin

yejiayu (changes requested):
I don't quite understand your logic.

It seems to me that you only need to collect metrics once each time a new task is inserted into the execution queue and once each time the task is executed.

homura:
Wouldn't it be clearer to just monitor the current `exec_height`?

I'm guessing that `executing_blocks` is trying to see how many blocks have been executed in a single execution, and is essentially monitoring how many blocks are currently behind, right? 

Therefore, I suggest that just monitoring `exec_height` is sufficient and we can build a module to monitor the current backward height

```
muta_consensus_commited_height - muta_consensus_executed_height
```

LycrusHamster (approved):

PR #434, opened 2020-08-27
feat(logger): add structured log api

zeroqn:
Add more structured logger api 

Fixes #

Ref #

No Breaking Change


LycrusHamster:
Yes indeed, you can definitely re-use target field for module.
trace_id and span_id would be Okay

PR #435, opened 2020-08-27
feat(logger): split log file by size

zeroqn:

This PR add ```file_size_limit``` to ```logger``` config. When log file reaches given file size limit, current log
file will be "archived" into ```[data].muta.[timestamp].log```.

Fixes #

Ref #
https://github.com/nervosnetwork/muta-docs/pull/62

No Breaking Change


LycrusHamster:
I doubt whether we should use the yml config by default, nest the log.yml inside of config.yml :)
I don't know if we use customized appender and log4rs would dynamicaly scan and register and load need .rs files ?

PR #436, opened 2020-08-28
feat: cli for maintance

LycrusHamster:

feat: cli for maintance

due to the need of operation in production env, capabilty of manipulation of 'cold' chain is urgent.

'cold' chain means that handle problems while the chain is off

function:

get/set block
get/set latest block
backup/restore
query, WIP

WIP

No Breaking Change


KaoImin:
Please remove the rocksDB files created by the unit test.

yejiayu (changes requested):
First review, about WAL.

core/consensus/src/wal.rs
```diff
@@ -129,6 +154,177 @@ impl SignedTxsWAL {
...
+            fs::create_dir_all(&path).expect("Failed to create wal directory");
+        }
+
+        ConsensusWal {
+            path: path.as_ref().to_path_buf(),
+        }
+    }
+
+    #[muta_apm::derive::tracing_span(kind = "consensus_wal")]
+    pub fn update_overlord_wal(&self, ctx: Context, info: Bytes) -> ProtocolResult<()> {
```
KaoImin, line 177: You should create an overlord wal file per height, and append the later info to the file. This can make the code brief.

src/.gitignore
```diff
@@ -0,0 +1,6 @@
+tests/data
+
+tests/save
+
+
+!tests/init/**/*
```
KaoImin, line 6: Move these to the .gitignore in the root path.

KaoImin (changes requested):
The way to test the cli to modify the DB data should be to write the data in the test, rather than upload the database file。

core/consensus/src/wal.rs
```diff
@@ -129,6 +154,177 @@ impl SignedTxsWAL {
     }
 }
 
+#[test]
+fn a() {}
```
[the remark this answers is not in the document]
LycrusHamster: removed

core/consensus/src/wal.rs
```diff
@@ -129,6 +154,177 @@ impl SignedTxsWAL {
...
+        }
+
+        // 2nd, write info into file
+        let check_sum = Hash::digest(info.clone());
+
+        let mut content = BytesMut::new();
+        content.put(check_sum.as_bytes());
+        content.put(info);
+
+        let (data_path, timestamp) = {
```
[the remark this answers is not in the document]
LycrusHamster: if some one copy the file to another place and restore it, modify time, access time and create time could be changed.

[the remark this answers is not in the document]
LycrusHamster: use another 'line' of file would be vunable.
If crash happens while update the file, all 'lines' may get wrong and WAL lose efficay.

core/consensus/src/wal.rs
LycrusHamster: No, that's useless because on each height, there could be more than one write operate on wal.
And modification of add height param of interface has to be done

src/.gitignore
LycrusHamster: Fine

core/consensus/src/wal.rs
```diff
@@ -129,6 +154,177 @@ impl SignedTxsWAL {
...
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .map_err(ConsensusError::SystemTime)?;
+
+                let timestamp = timestamp.as_millis();
+
+                let mut data_path = dir_path.clone();
+
+                data_path.push(timestamp.to_string());
+
+                // fixme
```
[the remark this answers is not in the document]
LycrusHamster: none, I forgot to remove this.

core/consensus/src/wal.rs
```diff
@@ -63,6 +68,26 @@ impl SignedTxsWAL {
         Ok(())
     }
 
+    pub fn available_height(&self) -> ProtocolResult<Vec<u64>> {
```
KaoImin, line 71: What does this function use？I have not seen it.

KaoImin:
The part of the overlord wal seems well.

core/consensus/src/wal.rs
LycrusHamster: oh~,that's for cli maintainance to list all wal file for their heights and help to remove wal files.
That's could be only used in cli maintainance tool.

core/consensus/src/tests/engine.rs
```diff
@@ -23,11 +23,12 @@ use crate::engine::ConsensusEngine;
 use crate::fixed_types::FixedPill;
 use crate::status::StatusAgent;
 use crate::util::OverlordCrypto;
-use crate::wal::SignedTxsWAL;
+use crate::wal::{ConsensusWal, SignedTxsWAL};
 
 use super::*;
 
-static FULL_TXS_PATH: &str = "./free-space/engine";
+static FULL_TXS_PATH: &str = "./free-space/engine/txs";
```
yejiayu, line 30: Save them all together in the `data_path` directory? This allows us to view all our data in one unified directory.

core/consensus/src/wal.rs
yejiayu: You should write a comment on this function for other developers to know what it does.

src/cli.rs
```diff
@@ -0,0 +1,593 @@
...
+                        .about("get block of [BLOCK_HEIGHT]"),
+                )
+                .subcommand(
+                    clap::SubCommand::with_name("set")
+                        .arg(clap::Arg::with_name("BLOCK").required(true))
+                        .about("upsert target block by [BLOCK], [BLOCK] is in JSON format"),
+                ),
+        )
+        .subcommand(
+            clap::SubCommand::with_name("wal")
```
yejiayu, line 113: Should remove both the consensus and the transaction `WAL`

src/cli.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    pub async fn latest_block_set(&self, height: u64) -> ProtocolResult<()> {
+        let last = self.storage.get_latest_block(Context::new()).await?;
+
+        let block = self.block_get(height).await?;
+        let block = match block {
+            Some(blk) => blk,
+            None => return Err(MainError::Other(height.to_string()).into()),
+        };
+        println!(
```
yejiayu, line 310: Use log?

src/cli.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    pub async fn latest_block_get(&self) -> ProtocolResult<Block> {
+        self.storage.get_latest_block(Context::new()).await
+    }
+
+    pub fn block(&self) -> ProtocolResult<()> {
+        let mut rt = tokio::runtime::Runtime::new().expect("new tokio runtime");
+
+        match self.matches.subcommand() {
+            ("set", Some(cmd)) => {
```
yejiayu, line 332: Does this mean we can replace a block arbitrarily and not do any data correction? This would disrupt the continuity of the blockchain.

yejiayu (changes requested):
About maintenance mode

1. I think it is possible to provide the ability to roll back the latest blocks, but users should not be allowed to modify any blocks.
2. Users should be forced to back up or prompted to back up before modifying because it's too dangerous.

src/cli.rs
LycrusHamster: Will remove 2 'clear' commands for consensus wal and txs wal and add a 'clear' command directly under 'wal'

src/cli.rs
LycrusHamster: excellent

core/consensus/src/tests/engine.rs
LycrusHamster: no, that will confuse

core/cli/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+[package]
+name = "cli"
+version = "0.2.0-rc.2.1"
+authors = ["LycrusHamster <liang2929@hotmail.com>"]
```
yejiayu, line 4:
```suggestion
authors = ["Muta Dev <muta@nervos.org>"]
```

LycrusHamster: done

KaoImin (approved):

PR #450, opened 2020-09-09
fix(logger): add structured api

zeroqn:
Add structured api document to logger README

Fixes #

Ref #

No Breaking Change

### DIFF
--- a/common/logger/README.md
+++ b/common/logger/README.md
@@ -60,6 +60,23 @@ This signature of the function is showed below. The `JsonValue` is a `enum` from
 pub fn metrics(name: &str, mut content: JsonValue)
 ```
 
+## Structured Event Log With TraceId Included
+
+Structured event log api provide a convenient way to log structured json data. It's signature is provided as below:
+
+```rust
+pub fn log(level: Level, module: &str, event: &str, ctx: &Context, mut msg: JsonValue)
+```
+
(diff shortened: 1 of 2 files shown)


common/logger/README.md
```diff
@@ -60,8 +60,25 @@ This signature of the function is showed below. The `JsonValue` is a `enum` from
...
+pub fn log(level: Level, module: &str, event: &str, ctx: &Context, mut msg: JsonValue)
+```
+
+`module` should be your component name, `event` is just event name, better begin with 4 chars with 4 digits
+to identify this event. `Context` is used to extract trace id. `msg` is `JsonValue` which is same as `metrics`.
+
+Useage example:
+
+```rust
+common_logger::log(Level::Info, "network", "netw0001", &ctx, json!{"music": "beautiful world"});
```
KaoImin, line 77: The return type of `json!()` is `Value`, however, the `log()` method needs `JsonValue`.

KaoImin (changes requested):

common/logger/README.md
zeroqn: Oh, ```common-logger``` uses ```json``` not ```serde_json```.
Replace ```json!``` with ```common_logger::object!```.

KaoImin (approved):

PR #458, opened 2020-09-16
perf: remove unnecessary clone

zeroqn:
Remove unnecessary clone, reduce memory and cpu usage. Especial remove ```Block``` clone in mempool.

Note:
This PR introduce new api to Storage, which save ```BlockHeader```.

Fixes #

Ref #

No Breaking Change

### DIFF
--- a/protocol/src/types/primitive.rs
+++ b/protocol/src/types/primitive.rs
@@ -15,6 +15,7 @@ use ophelia_secp256k1::Secp256k1PublicKey;
 use serde::de;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use smol_str::SmolStr;
 
 use crate::fixed_codec::{FixedCodec, FixedCodecError};
 use crate::types::TypesError;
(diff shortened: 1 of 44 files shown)


KaoImin (approved):

protocol/src/types/primitive.rs
```diff
@@ -24,19 +25,23 @@ pub const METADATA_KEY: &str = "metadata";
...
+pub fn address_hrp() -> SmolStr {
+    ADDRESS_HRP.load().as_ref().clone()
 }
 
 pub fn init_address_hrp(address_hrp: String) {
     if ADDRESS_HRP_INITED.load(Ordering::SeqCst) {
         panic!("address hrp can only be inited once");
     }
 
+    if address_hrp.as_bytes().len() > 21 {
```
LycrusHamster, line 41: please use is_heap_allocated() instead of 21

And, do not restrict the bech32 standard
I reckon if the hrp is allocated on heap, just throw a warning of perf impacting

KaoImin (approved):

PR #459, opened 2020-09-18
feat(consensus): metric proposal ordered tx len and sync tx len

KaoImin:
Metric proposal ordered transaction length and sync transaction length.

Fixes #

Ref #

No Breaking Change

### DIFF
--- a/common/apm/src/metrics/consensus.rs
+++ b/common/apm/src/metrics/consensus.rs
@@ -72,6 +72,10 @@ lazy_static! {
         "The committed transactions"
     )
     .unwrap();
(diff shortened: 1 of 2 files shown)


common/apm/src/metrics/consensus.rs
```diff
@@ -72,6 +72,10 @@ lazy_static! {
         "The committed transactions"
     )
     .unwrap();
+    pub static ref ENGINE_ORDER_TX_GAUGE: IntGauge =
```
homura, line 75: Would you explain the purpose of this gauge? I'm guessing you want to be accumulating the number of `proposals_hash`s and `transaction`s received from the proposal.

rev-chaos (approved):

